PR #5, opened 2020-02-11
Add Project API

rafael-azevedo:
This PR adds the project API.  To do so I moved common types between the project and projectClaim types to a common files. I also added updated examples of both types.


rafael-azevedo:
When testing this implementation I realized that Kube has a default type called projects. While this implementation does work you can get the object by saying `oc get projects.gcp.managed.openshift.io` we many want to rethink the name of this api for more clarity .

NautiluX:
I'm in favor of renaming. GCPProject maybe?

jhernand:
@tzvatot please review.

deploy/crds/gcp_v1alpha1_project_crd.yaml
```diff
@@ -0,0 +1,115 @@
...
+  - JSONPath: .status.state
+    description: Status of the project
+    name: State
+    type: string
+  - JSONPath: .spec.projectClaimCRLink.name
+    description: Name of corresponding project claim CR
+    name: ClaimName
+    type: string
+  - JSONPath: .spec.projectClaimCRLink.namespace
+    description: Namesspace of corresponding project CR
```
NautiluX: should this be project claim CR?

pkg/apis/gcp/v1alpha1/common_types.go
```diff
@@ -0,0 +1,13 @@
+package v1alpha1
+
+// LegalEntity contains Red Hat specific identifiers to the original creator the clusters
+type LegalEntity struct {
+	Name string `json:"name"`
+	ID   string `json:"id"`
+}
+
+// NamespacedName contains the name of a secret and its namespace
```
NautiluX: We use this as generic name of any Kubernetes object, right? Like, as a reference to a `ProjectClaim`, so I assume it's not just secret.

pkg/apis/gcp/v1alpha1/project_types.go
```diff
@@ -0,0 +1,91 @@
+package v1alpha1
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
```
NautiluX: Shouldn't we remove this comment?

rafael-azevedo: We have left this in on other repos, but I don't see why we shouldn't remove it.  Do you think we should just remove it ?

rafael-azevedo:
Updated the type to name `GCPProject` and changed the `GCPProjectName` field in both api types to `GCPProjectID` for better ease of reading. The `GCPProjectID` is the name of the Project have have instantiated in the GCP cloud and not the CR name.

pkg/apis/gcp/v1alpha1/project_types.go
NautiluX: I understand it as note by framework for us to remove once we understood this file is for us to touch ;)

rafael-azevedo: I will remove on all files in this operator

PR #7, opened 2020-02-18
Generate project id

NautiluX:
This PR introduces generated controller code and sets a unique ID as `GCPProjectID` in the `ProjectReference` CR.
Before merging, let's discuss the generation of project IDs, currently it uses plain [GUIDs](https://tools.ietf.org/html/rfc4122).
Ensuring a project actually doesn't exist in GCP is out of scope of this PR, but shouldn't be complicated include, when first GCP logic comes in.


pkg/controller/projectreference/projectreference_controller.go
```diff
@@ -0,0 +1,116 @@
...
+		return err
+	}
+
+	// Watch for changes to primary resource ProjectReference
+	err = c.Watch(&source.Kind{Type: &gcpv1alpha1.ProjectReference{}}, &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return err
+	}
+
+	// TODO(user): Modify this to be the types you create that are owned by the primary resource
```
rafael-azevedo: we can remove this Block that watches pod resources.

pkg/controller/projectreference/projectreference_controller.go
```diff
@@ -0,0 +1,116 @@
...
+
+func (r *ReconcileProjectReference) getProjectReference(request reconcile.Request) (*gcpv1alpha1.ProjectReference, error) {
+	instance := &gcpv1alpha1.ProjectReference{}
+	err := r.client.Get(context.TODO(), request.NamespacedName, instance)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// Request object not found, could have been deleted after reconcile request.
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue
+			return nil, nil
```
rafael-azevedo: Instead of returning nil  we should return an empty pointer or object. 

Return` &gcpv1alpha1.ProjectReference{}`

pkg/controller/projectreference/projectreference_controller.go
```diff
@@ -0,0 +1,116 @@
...
+// The Controller will requeue the Request to be processed again if the returned error is non-nil or
+// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+func (r *ReconcileProjectReference) Reconcile(request reconcile.Request) (result reconcile.Result, err error) {
+	result = reconcile.Result{}
+	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	reqLogger.Info("Reconciling ProjectReference")
+
+	instance, err := r.getProjectReference(request)
+	if (instance == nil) || (err != nil) {
+		return
```
rafael-azevedo: return is missing objects

pkg/controller/projectreference/projectreference_controller.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	if instance.Spec.GCPProjectID == "" {
+		err = r.updateProjectName(instance) //this will update the spec, resulting in another run in the loop, hence return
+		return
```
rafael-azevedo: return is missing objects

pkg/controller/projectreference/projectreference_controller.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	return
```
rafael-azevedo: return is missing objects

pkg/controller/projectreference/projectreference_controller.go
NautiluX: wanted to do that, but can't compare two of those with `==`

pkg/controller/projectreference/projectreference_controller.go
NautiluX: the return values are named in the function header, if we find this confusing I can change it.

pkg/controller/projectreference/projectreference_controller.go
NautiluX: see above

pkg/controller/projectreference/projectreference_controller.go
NautiluX: see above

pkg/controller/projectreference/projectreference_controller.go
NautiluX: changed to using cmp.Equal, which seems to be able to compare those.

PR #12, opened 2020-02-20
OSD-2888: get orgParentFolderID from configmap

boranx:


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -33,11 +33,11 @@ const (
 	// clusterPlatformLabel is the label on a cluster deployment which indicates whether or not a cluster is on GCP platform
 	clusterPlatformLabel = "hive.openshift.io/cluster-platform"
 	clusterPlatformGCP   = "gcp"
-	orgParentFolderID    = "240634451310" // Service Delivery org subfolder
 
 	// secret information
 	gcpSecretName         = "gcp"
 	orgGcpSecretName      = "gcp-project-operator"
+	orgGcpConfigMap       = "gcp-project-operator"
```
NautiluX: I think all the other variables are related to gcp secrets, and to avoid confusion they are prefixed with "org". This is not the case for the config map so maybe move it to the "Operator config" block above (line 27)

pkg/util/util.go
```diff
@@ -73,6 +72,28 @@ func NewGCPSecretCR(namespace, creds string) *corev1.Secret {
 	}
 }
 
+// GetGCPParentFolderFromConfigMap returns a configmap value if it's exist
```
NautiluX: typo: "If it exists"

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -182,6 +182,13 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (recon
 		return reconcile.Result{}, err
 	}
 
+	// Get org orgParentFolderID from configmap, use default one if not exist
```
NautiluX: typo: "use default one if it doesn't exist"

pkg/util/util.go
```diff
@@ -73,6 +72,28 @@ func NewGCPSecretCR(namespace, creds string) *corev1.Secret {
...
+	if err != nil {
+		return orgParentFolderID, fmt.Errorf("clusterdeployment.GetGCPParentFolderFromConfigMap.Get %v", err)
+	}
+	orgParentFolderIDconfig, ok := configmap.Data["orgParentFolderID"]
+	if !ok {
+		return orgParentFolderID, fmt.Errorf("GCP config %v did not contain key %v",
+			name, "orgParentFolderID")
+	}
+
+	return orgParentFolderIDconfig, nil
```
NautiluX: For now we only need the parent folder, so this function is ok for now, but sooner or later we will read more values from that config map and then we will want a config struct to be filled. If you want, I think you can already create it and return it here instead of the single value.

pkg/controller/clusterdeployment/clusterdeploymet_controller_test.go
```diff
@@ -70,6 +70,17 @@ func TestReconcile(t *testing.T) {
...
+		{
+			name:        "get orgParentFolderID from configmap",
+			expectedErr: nil,
+			localObjects: []runtime.Object{
+				builders.NewTestClusterDeploymentBuilder().GetClusterDeployment(),
+				builders.NewTestConfigMapBuilder(orgGcpSecretName, operatorNamespace, "111111").GetConfigMap(),
+				builders.NewTestSecretBuilder(orgGcpSecretName, operatorNamespace, "testCreds").GetTestSecret(),
+				builders.NewTestSecretBuilder(gcpSecretName, testNamespace, "testCreds").GetTestSecret(),
+			},
+			setupGCPMock: func(r *mockGCP.MockClientMockRecorder) { gomock.Any() },
```
NautiluX: I have no idea what this does, and how this tests your code. I.e. where does it test what you put into the `ConfigMap` is actually digged out by `GetGCPParentFolderFromConfigMap`?

pkg/util/util.go
boranx: @NautiluX makes sense, I'll modify here

pkg/controller/clusterdeployment/clusterdeploymet_controller_test.go
boranx: this test is fixed for testing negative scenario: if the configmap does not contain orgParentFolderID

pkg/util/util.go
```diff
@@ -119,6 +118,59 @@ func GetBillingAccountFromSecret(kubeClient kubeclientpkg.Client, namespace, nam
...
+func ConfigMapExist(kubeClient kubeclientpkg.Client, name, namespace string) bool {
+	c := &corev1.ConfigMap{}
+	if err := kubeClient.Get(context.TODO(), kubetypes.NamespacedName{Name: name, Namespace: namespace}, c); err != nil {
+		return false
+	}
+
+	return true
+}
+
+// getConfigMap returns a configmap that should contain orgParentFolderID.
```
NautiluX: this method has nothing to do with the ParentFolderID, I think it's not necessary to explain that in the comment

pkg/util/util.go
```diff
@@ -119,6 +118,59 @@ func GetBillingAccountFromSecret(kubeClient kubeclientpkg.Client, namespace, nam
 	return billingAccount, nil
 }
 
+// NewConfigMapCR creates a ConfigMap in specified namespace with default orgParentFolderID value
+func NewConfigMapCR(name, namespace, orgParentFolderID string) *corev1.ConfigMap {
```
NautiluX: is a ConfigMap a CR? I don't think so, so maybe the method should be called differently

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -166,6 +170,19 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (recon
...
+	if util.ConfigMapExist(r.client, operatorNamespace, orgGcpConfigMap) {
+		orgParentFolderID, err = util.GetGCPParentFolderFromConfigMap(r.client, operatorNamespace, orgGcpConfigMap)
+		if err != nil {
+			reqLogger.Error(err, "could not get orgParentFolderID from the ConfigMap:", orgGcpConfigMap, "Operator Namespace", operatorNamespace)
+			return reconcile.Result{}, err
+		}
+	} else {
+		configmap := util.NewConfigMapCR(orgGcpConfigMap, operatorNamespace, orgParentFolderID)
+		reqLogger.Info(fmt.Sprintf("%s has been created in Namespace: %s :: Moving with default orgParentFolderID", configmap.GetName(), operatorNamespace))
+	}
```
NautiluX: As discussed, we can just return an error here, maybe even increase the time when this Project will be reconciled again, I think this is possible using reconcile.Result.RequeueAfter.  On a side note, please consider putting this into its own method.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -23,6 +23,11 @@ import (
 
 var log = logf.Log.WithName("controller_clusterdeployment")
 
+// Configmap related configs
+const orgGcpConfigMap = "gcp-project-operator"
+
+var orgParentFolderID = "240634451310"
```
NautiluX: should be `""`, not a default value

pkg/util/util.go
boranx: CR is removed now.

pkg/util/util.go
boranx: that part is removed.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
boranx: It's now created inside reconcile loop.

NautiluX:
FYI @rafael-azevedo you may need to rebase to this once merged.

PR #14, opened 2020-02-25
OSD-2924: Migrate Billing account to configmap

boranx:


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -174,7 +174,8 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (recon
 		return reconcile.Result{}, nil
 	}
 
-	orgParentFolderID, err := util.GetGCPParentFolderFromConfigMap(r.client, operatorNamespace, orgGcpConfigMap)
+	configmap := util.GetConfigMapOperations(r.client, orgGcpConfigMap, operatorNamespace)
```
NautiluX: I find the name ConfigMapOperations confusing when first reading it... It's just a wrapper around the config map right? Shouldn't we call it `GcpProjectOperatorConfigMap` or (less verbose) `OperatorConfig`?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -211,9 +212,9 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (recon
 	}
 
 	// TODO(Raf) Set quotas
-	billingAccount, err := util.GetBillingAccountFromSecret(r.client, operatorNamespace, orgGcpSecretName)
+	billingAccount, err := configmap.GetBillingAccount()
 	if err != nil {
-		reqLogger.Error(err, "could not get org billingAccount from secret", "Secret Name", orgGcpSecretName, "Operator Namespace", operatorNamespace)
+		reqLogger.Error(err, "could not get org billingAccount from configmap", "Configmap", orgGcpConfigMap, "Operator Namespace", operatorNamespace)
```
NautiluX: what does the `org` prefix of the variable `orgGcpConfigMap` mean?

pkg/util/util.go
```diff
@@ -30,6 +30,70 @@ const (
 	osdServiceAccountName = "osd-managed-admin"
 )
 
+// ConfigMapOperations store informations required for operations
+type ConfigMapOperations struct {
```
NautiluX: if this is its own type with methods, I'd prefer to move it into its own file, potentially outside of the util folder.

pkg/util/util.go
```diff
@@ -30,6 +30,70 @@ const (
...
+			c.name, key)
+	}
+
+	return value, nil
+}
+
+// GetParentFolder returns orgParentFolderID if the key exists in configmap
+func (c *ConfigMapOperations) GetParentFolder() (string, error) {
+	value, err := c.getValue("orgParentFolderID")
+	if err != nil {
```
NautiluX: I don't think you need to wrap that error here again... you could just return `value, err`

pkg/util/util.go
```diff
@@ -30,6 +30,70 @@ const (
...
+		return "", fmt.Errorf("configmap operations failed: %v", err)
+	}
+
+	return value, nil
+}
+
+// GetBillingAccount returns billingaccount if the key exists in configmap
+func (c *ConfigMapOperations) GetBillingAccount() (string, error) {
+	value, err := c.getValue("billingaccount")
+	if err != nil {
```
NautiluX: same as above `return value, err`

pkg/controller/clusterdeployment/clusterdeployment_controller.go
rafael-azevedo: maybe `orgGCPConfigMap` would be better as `ConfigMapName` or something like that

pkg/util/util.go
rafael-azevedo: +1

pkg/controller/clusterdeployment/clusterdeployment_controller.go
boranx: replaced with `GcpProjectOperatorConfigMap`

pkg/controller/clusterdeployment/clusterdeployment_controller.go
boranx: replaced with `ConfigMapName`

pkg/util/util.go
boranx: migrated to `configmap` folder. Can be renamed again, not sure that suits to here well

pkg/util/util.go
boranx: done

pkg/util/util.go
boranx: done

mmazur:
Please make sure this PR ends up containing an update to `README.md` in the "GCP Secret" section, that reflects the new setup.

boranx:
@mmazur @NautiluX what do you think about the current state of the `README`?

rafael-azevedo:
/lgtm 

We will need to wait until integration migration is done to merge this so that we can make sure all changes are working.

pkg/configmap/configmap.go
```diff
@@ -0,0 +1,66 @@
...
+	"context"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	kubetypes "k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// GcpProjectOperatorConfigMap store informations required for operations
+type GcpProjectOperatorConfigMap struct {
```
rafael-azevedo: @boranx This struct should reflect that data we are pulling from the configMap so in this case 

```
OperatorConfigMap struct {
BillingAccount string 'json:"billingaccount"`
ParentFolderID string 'json:"parentfolderid"
}
```

This way we can have a function the gets the configmap and unmarshalls it into this object and we can pass this object around our code. As we add stuff to the configmap all we need to do is add it to this object. 


so we would unmarshall `configmap.Data` later right into this struct

I also think we should change the variables to the following 

```
"parentFolderID" // Since now we are not longer using the ORG root ID ,but a different folder for each env
"billingAccount" // just formatting so we are consistent 
```

rafael-azevedo: we should also move 

`const orgGcpConfigMap = "gcp-project-operator"`

to the new configmap file and maybe `OperatorConfigMapName` would be better and the configmap data struct can be `OperatorConfigMap`

boranx:
Since this became too dirty, I'll create a new PR that contains requested changes

PR #15, opened 2020-02-25
ProjectClaim controller creates ProjectReference

NautiluX:


rafael-azevedo:
The only thing that is confusing is the naming convention of the "ProjectClaimReconciler" since there is already another reconcile object. Though I am not sure on a good name.

rafael-azevedo:
Have you also tested this workflow locally  ?

NautiluX:
Thanks for your review @rafael-azevedo, I'll try to come up with a better name. As discussed I tested the workflow locally and will have another look to make sure no outdated objects are used.

PR #17, opened 2020-02-28
Ensure project id meets gcp requirements

NautiluX:
OSD-3007


NautiluX:
Thanks for pointing this out. I added tests for all the requirements and changed to a different generation:
- generate v4 GUID
- sha1 hash GUID
- take first 26 bytes as characters
- add `osd-` prefix to ensure it starts with a letter

This should still have a sufficient low probability for collisions, while being compliant to the naming requirements from GCP.

pkg/controller/projectreference/projectreference_controller_test.go
```diff
@@ -83,6 +83,30 @@ var _ = Describe("ProjectReference controller reconcilation", func() {
 			})
 		})
 
+		Context("generated project names", func() {
+			It("are shorter than 30 characters", func() {
+				projectName := generateProjectName()
+				Expect(len(projectName)).To(BeNumerically("<=", 30))
+			})
+
+			It("are shorter than 6 characters", func() {
```
drpaneas:
```suggestion
			It("are longer than 6 characters", func() {
```

PR #19, opened 2020-03-03
Project Reference

rafael-azevedo:
This PR migrates logic from the ClusterDeployment controller to the ProjectReference controller.


pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -0,0 +1,343 @@
...
+		policy:   policy,
+	}, nil
+}
+
+// SetIAMPolicy attempts to update policy if the policy needs to be modified
+func (r *ReferenceAdapater) SetIAMPolicy(serviceAccountEmail string) error {
+	// Checking if policy needs to be updated
+	var retry int
+	for {
+		retry++
```
NautiluX: why not use a boring `for retry = 0; retry <4; retry++` loop?

pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -0,0 +1,343 @@
...
+			r.logger.Info("Project lifecycleState == ACTIVE")
+			return nil
+		case "DELETE_REQUESTED":
+			return operrors.ErrInactiveProject
+		default:
+			r.logger.Error(operrors.ErrUnexpectedLifecycleState, "Unexpected LifecycleState", project.LifecycleState)
+			return operrors.ErrUnexpectedLifecycleState
+
+		}
+		return nil
```
drpaneas: It is unreachable code because you have individual `return` statement for each case.

```suggestion

```

pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -0,0 +1,343 @@
...
+	"serviceusage.googleapis.com",
+	"cloudresourcemanager.googleapis.com",
+	"storage-component.googleapis.com",
+	"storage-api.googleapis.com",
+	"dns.googleapis.com",
+	"iam.googleapis.com",
+	"compute.googleapis.com",
+}
+
+var OSDRequiredRoles = []string{
```
drpaneas: Linter-wise, since it starts with uppercase, it would be exported, and we need a comment/description.

```suggestion
// OSDRequiredRoles blah blah blah
var OSDRequiredRoles = []string{
```

Replace my _blahblahblah_ with the purpose its fits (I guess the reply to my question above, would fit for a description).

drpaneas: Also, according to the [Required GCP Permissions](OSDRequiredRoles) there is a mismatch. In some case you use `iam.` while the docs are using `compute.`. Have a look at the link and see if this OK.

NautiluX:
The build is failing btw.
```
# github.com/openshift/gcp-project-operator/pkg/controller/projectreference
pkg/controller/projectreference/projectreference_adapter.go:173:25: undefined: util.GetBillingAccountFromSecret
pkg/controller/projectreference/projectreference_controller.go:207:25: undefined: util.GetGCPParentFolderFromConfigMap
# github.com/openshift/gcp-project-operator/pkg/controller/projectreference
vet: pkg/controller/projectreference/projectreference_adapter.go:173:30: GetBillingAccountFromSecret not declared by package util
make: *** [standard.mk:65: gocheck] Error 2
```

pkg/controller/projectreference/projectreference_controller.go
```diff
@@ -72,34 +88,189 @@ func (r *ReconcileProjectReference) Reconcile(request reconcile.Request) (reconc
...
+		return r.requeueOnErr(err)
+	}
+
+	// Make projectReference  be processed based on state of ProjectClaim and Project Reference
+	switch {
+	// If we are in a creating state break from the loop and conitnue to process CR
+	case projectReference.Status.State == gcpv1alpha1.ProjectReferenceStatusCreating:
+		break
+	// If projectReference is in an unrecoverable error state do not process
+	case projectReference.Status.State == gcpv1alpha1.ProjectReferenceStatusReady && adapter.projectClaim.Status.State == gcpv1alpha1.ClaimStatusVerification:
```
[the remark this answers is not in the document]
rafael-azevedo: As we talked about. I will be removing the verification state. In the future the Project Reference will set the Project Claim to a ready state directly.

pkg/controller/projectreference/projectreference_controller.go
```diff
@@ -72,34 +88,189 @@ func (r *ReconcileProjectReference) Reconcile(request reconcile.Request) (reconc
...
+				reqLogger.Error(err, "Error updating ProjectReference Status")
+				return r.requeueOnErr(err)
+			}
+		}
+		reqLogger.Error(err, "Could not create ProjectID")
+		return r.requeueOnErr(err)
+	}
+
+	reqLogger.Info("Configuring APIS")
+	// Set condition billing has been created and skip that this step if condition is true
```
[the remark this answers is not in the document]
rafael-azevedo: This comment was suppose to be a "TODO()" I will adjsut this comment

pkg/controller/projectreference/projectreference_controller.go
```diff
@@ -72,34 +88,189 @@ func (r *ReconcileProjectReference) Reconcile(request reconcile.Request) (reconc
...
+}
+
+func (r *ReconcileProjectReference) requeue() (reconcile.Result, error) {
+	return reconcile.Result{Requeue: true}, nil
 }
 
-func (r *ReconcileProjectReference) updateProjectName(instance *gcpv1alpha1.ProjectReference) error {
-	instance.Spec.GCPProjectID = uuid.New().String()
-	return r.client.Update(context.TODO(), instance)
+func (r *ReconcileProjectReference) requeueAfter(duration time.Duration) (reconcile.Result, error) {
```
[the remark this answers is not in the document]
rafael-azevedo: Yes , this would be a good idea. I will pass the error through this function.

pkg/controller/projectreference/projectreference_controller.go
```diff
@@ -72,34 +88,176 @@ func (r *ReconcileProjectReference) Reconcile(request reconcile.Request) (reconc
...
+		return r.doNotRequeue()
 	}
-	return instance, err
+
+	// only make changes to ProjectReference if ProjelctClaim is pending
+	// if adapter.projectClaim.Status.State != gcpv1alpha1.ClaimStatusPending {
+	// 	return r.requeueAfter(5 * time.Second)
+	// }
+
+	// make sure we meet mimimum requiresments to process request and set its state to creating or error if its not supported
```
NautiluX: typo requiresments

pkg/controller/projectreference/projectreference_controller.go
```diff
@@ -72,34 +88,176 @@ func (r *ReconcileProjectReference) Reconcile(request reconcile.Request) (reconc
...
+	}
+
+	if projectReference.Spec.GCPProjectID == "" {
+		reqLogger.Info("Creating ProjectID in ProjectReference CR")
+		err := adapter.updateProjectID()
+		if err != nil {
+			reqLogger.Error(err, "Could not update ProjectID in Project Reference CR")
+			return r.requeueOnErr(err)
+		}
+		return r.requeue()
```
NautiluX:
```suggestion
	return r.requeueOnErr(err)
```
Shouldn't this be just requeueOnErr? Since updateProjectID will change the ProjectReference, it will be reconciled again anyhow.

pkg/util/util.go
```diff
@@ -72,6 +72,24 @@ func NewGCPSecretCR(namespace, creds string) *corev1.Secret {
 	}
 }
 
+// NewGCPSecretCRV2 returns a Secret CR formatted for GCP for use in projectreference controller until refactor
```
NautiluX: I don't fully understand until when this function will be used and which refactoring is needed here, can you detail that out a bit?

pkg/controller/projectreference/projectreference_controller.go
```diff
@@ -72,34 +88,176 @@ func (r *ReconcileProjectReference) Reconcile(request reconcile.Request) (reconc
...
+	reqLogger.Info("Creating Credentials")
+	err = adapter.createCredentials()
+	if err != nil {
+		reqLogger.Error(err, "Error creating credentials")
+		return r.requeueAfter(5*time.Second, err)
+	}
+
+	log.Info("Setting Status on projectReference")
+	projectReference.Status.State = gcpv1alpha1.ProjectReferenceStatusReady
+	err = r.client.Status().Update(context.TODO(), projectReference)
```
NautiluX: Not sure what happens on a call of `Update()` if nothing changed, hopefully it doesn't result in reconciling the ProjctClaim again. Maybe it makes sense to explicitly check this before?

pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -0,0 +1,341 @@
...
+			return err
+		}
+		return err
+	}
+
+	return nil
+}
+
+func (r *ReferenceAdapter) configureAPIS() error {
+	billingAccount, err := util.GetBillingAccountFromSecret(r.kubeClient, operatorNamespace, orgGcpSecretName)
```
mmazur: This needs to be taken from configmap before merge, per current HEAD.

pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -0,0 +1,341 @@
...
+
+	r.logger.Info("Enabling Billing API")
+	err = r.gcpClient.EnableAPI(r.projectReference.Spec.GCPProjectID, "cloudbilling.googleapis.com")
+	if err != nil {
+		r.logger.Error(err, fmt.Sprintf("Error enabling %s api for project %s", "cloudbilling.googleapis.com", r.projectReference.Spec.GCPProjectID))
+		return err
+	}
+
+	r.logger.Info("Linking Cloud Billing Account")
+	// https://groups.google.com/forum/#!topic/gce-discussion/K_x9E0VIckk
```
mmazur: The comment can be dropped, this works without issues if proper perms are given to the service account.

pkg/controller/projectreference/projectreference_controller.go
rafael-azevedo: @NautiluX This is calling Update specifically on the Status since we are updating it to ready.

rafael-azevedo: and it does requeue. If you run the code you will see that on the next reconcile it will be caught at line 138 and then it will update the ProjectClaim status to ready as well.

pkg/util/util.go
rafael-azevedo: Since clusterdeployment controller is still in place we need to keep the original function `NewGCPSecretCR` intact since I do not want to modify that code, Since in the new code the ProjectClaim specifies the Namespaces and Name of the secret to be created I added namespacedName which contains the name and namespace to a `v2` version of this function. Once we remove the clusterdeployment controller, we will be able to delete the original function and rename this one.

pkg/controller/projectreference/projectreference_adapter.go
rafael-azevedo: This is a good point. For the time being since I have already tested and created testing for the codebase as is. I added the this function back to the util package. We should make a separate card to remove these from the ProjectReference controller. Since there will be a few items refactored around the new configmap.

NautiluX: I think the billing account doesn't exist anymore in the secret since we switched to the service accounts + config map

pkg/util/util.go
NautiluX: Understood. Can you put that to the comment? :D

pkg/controller/projectreference/projectreference_adapter.go
mmazur: In that case I need to recreate the secret on int with billing account re–added, before merge. I'll add it to my todo.

NautiluX: See https://github.com/rafael-azevedo/gcp-project-operator-1/pull/1

pkg/util/util.go
NautiluX: See https://github.com/rafael-azevedo/gcp-project-operator-1/pull/1

pkg/controller/projectreference/projectreference_adapter.go
NautiluX: See https://github.com/rafael-azevedo/gcp-project-operator-1/pull/1

PR #23, opened 2020-03-09
Added ProjectID to ProjectClaim when Project Reference is READY

rafael-azevedo:
This PR updates the ProjectClaim with the GCPProjectID when the Project Reference is Ready.


rafael-azevedo:
> Would love to see a test for this, though ;)

I'll add test :)

PR #24, opened 2020-03-09
Use vendored dependencies in gobuild

NautiluX:
* add --mod=vendor to gobuild target


NautiluX:
It's still downloading things... let me see

NautiluX:
`go vet`, `gofmt` and `go test` seem to ignore that flag. But at least the build should fail now if you forget to `go mod vendor` before committing.

PR #28, opened 2020-03-11
ProjectClaim controller maintains ProjectClaim state

ninabauer:
Changes ProjectClaim state to Pending when processing is started, and to PendingProject when processing of a ProjectReference is ready to begin. Added mock StatusWriter for testing.


pkg/controller/projectclaim/customresourceadapter.go
```diff
@@ -129,3 +129,29 @@ func (c *CustomResourceAdapter) EnsureProjectReferenceExists() error {
...
 }
+
+func (c *CustomResourceAdapter) EnsureProjectClaimState(state gcpv1alpha1.ClaimStatus) error {
+	if c.projectClaim.Status.State == state {
+		return nil
+	}
+
+	if state == gcpv1alpha1.ClaimStatusPending {
+		if c.projectClaim.Status.State != "" {
+			return nil
```
NautiluX: Maybe we should return an error here as a unsupported state transition was requested.

pkg/controller/projectclaim/customresourceadapter.go
```diff
@@ -129,3 +129,29 @@ func (c *CustomResourceAdapter) EnsureProjectReferenceExists() error {
...
+		}
+	}
+
+	if state == gcpv1alpha1.ClaimStatusPendingProject {
+		if c.projectClaim.Status.State != gcpv1alpha1.ClaimStatusPending {
+			return nil
```
NautiluX: same here

pkg/controller/projectclaim/customresourceadapter_test.go
```diff
@@ -208,4 +210,74 @@ var _ = Describe("Customresourceadapter", func() {
...
+			BeforeEach(func() {
+				requestedState = gcpv1alpha1.ClaimStatusPending
+			})
+
+			Context("when ProjectClaim state is not empty", func() {
+				BeforeEach(func() {
+					currentState = gcpv1alpha1.ClaimStatusReady
+				})
+				It("doesn't change the ProjectClaim state", func() {
+					adapter.EnsureProjectClaimState(requestedState)
```
NautiluX:
```suggestion
					err := adapter.EnsureProjectClaimState(requestedState)
					Expect(err).To(HaveOccurred())
```
if  we change the function to return an error, we could verify here.

pkg/controller/projectclaim/customresourceadapter.go
NautiluX: need to allow ClaimStatusPendingProject here, though

pkg/controller/projectclaim/customresourceadapter.go
```diff
@@ -137,13 +138,13 @@ func (c *CustomResourceAdapter) EnsureProjectClaimState(state gcpv1alpha1.ClaimS
 
 	if state == gcpv1alpha1.ClaimStatusPending {
 		if c.projectClaim.Status.State != "" {
-			return nil
+			return fmt.Errorf("Unexpected ProjectClaim state transition: %s -> %s", c.projectClaim.Status.State, state)
```
NautiluX: As discussed, there may be occasions where state is `ClaimStatusPendingProject`, which we need to return `nil` for.

PR #36, opened 2020-03-17
GCP Project cleanup

drpaneas:
* Adds a finalizer to the ProjectReference CR during its creation
* Checks for DeletionTimestamp request and signals the cleanup
* Checks the lifecycle status of the GCP Project
* Deletes the Project from Google GCP or skip if it's already deleted
* Implements GetProject()
* Checks the exitence of secret and gets it
* Deletes the secret from the cluster or skip if it's already deleted
* Deletes the finalizer if cleanup done properly

Fixes OSD-3064

Tests have been added as well

### DIFF
--- a/pkg/controller/projectreference/projectreference_adapter.go
+++ b/pkg/controller/projectreference/projectreference_adapter.go
@@ -10,10 +10,11 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/google/uuid"
+	clusterapi "github.com/openshift/cluster-api/pkg/util"
 	gcpv1alpha1 "github.com/openshift/gcp-project-operator/pkg/apis/gcp/v1alpha1"
 	"github.com/openshift/gcp-project-operator/pkg/configmap"
 	"github.com/openshift/gcp-project-operator/pkg/gcpclient"
-	"github.com/openshift/gcp-project-operator/pkg/util"
+	gcputil "github.com/openshift/gcp-project-operator/pkg/util"
 	operrors "github.com/openshift/gcp-project-operator/pkg/util/errors"
 	"google.golang.org/api/cloudresourcemanager/v1"
 	"google.golang.org/api/googleapi"
@@ -24,6 +25,7 @@ import (
 
 const (
 	osdServiceAccountName = "osd-managed-admin"
+	finalizerName         = "finalizer.gcp.managed.openshift.io"
 )
 
 // OSDRequiredAPIS is list of API's, required to setup
@@ -205,6 +207,52 @@ func (r *ReferenceAdapter) updateProjectID() error {
 	return r.kubeClient.Update(context.TODO(), r.projectReference)
 }
 
+// IsDeletionRequested checks the metadata.deletionTimestamp of ProjectReference instance, and returns if delete requested.
+// The controllers watching the ProjectReference use this as a signal to know when to execute the finalizer.
+func (r *ReferenceAdapter) IsDeletionRequested() bool {
+	return r.projectReference.DeletionTimestamp != nil
+}
+
+// EnsureFinalizerAdded parses the meta.Finalizers of ProjectReference instance and adds finalizerName if not found.
+func (r *ReferenceAdapter) EnsureFinalizerAdded() error {
+	if !clusterapi.Contains(r.projectReference.GetFinalizers(), finalizerName) {
+		r.projectReference.SetFinalizers(append(r.projectReference.GetFinalizers(), finalizerName))
+		return r.kubeClient.Update(context.TODO(), r.projectReference)
+	}
+	return nil
+}
+
+// EnsureFinalizerDeleted parses the meta.Finalizers of ProjectReference instance and removes finalizerName if found;
+func (r *ReferenceAdapter) EnsureFinalizerDeleted() error {
+	r.logger.Info("Deleting Finalizer")
+	finalizers := r.projectReference.GetFinalizers()
+	if clusterapi.Contains(finalizers, finalizerName) {
+		r.projectReference.SetFinalizers(clusterapi.Filter(finalizers, finalizerName))
+		return r.kubeClient.Update(context.TODO(), r.projectReference)
+	}
+	return nil
+}
+
+// EnsureProjectCleanedUp deletes the project, the secret and the finalizer if they still exist
+func (r *ReferenceAdapter) EnsureProjectCleanedUp() error {
+	err := r.deleteProject()
+	if err != nil {
+		return err
+	}
+
+	err = r.deleteCredentials()
+	if err != nil {
+		return err
+	}
+
+	err = r.EnsureFinalizerDeleted()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
 func GenerateProjectID() (string, error) {
 	guid := uuid.New().String()
 	hashing := sha1.New()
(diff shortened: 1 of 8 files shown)


pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -152,6 +154,33 @@ func (r *ReferenceAdapter) updateProjectID() error {
...
 
+// IsDeletionRequested checks the metadata.deletionTimestamp of ProjectReference instance, and returns if delete requested.
+// The controllers watching the ProjectReference use this as a signal to know when to execute the finalizer.
+func (r *ReferenceAdapter) IsDeletionRequested() bool {
+	return r.projectReference.DeletionTimestamp != nil
+}
+
+// AddPrjRefFinalizer parses the meta.Finalizers of ProjectReference instance and adds prjRefFinalizer if not found.
+// It returns a modification to the ProjectReference instance in the cluster for the controller to handle the reconciliation.
+func (r *ReferenceAdapter) AddPrjRefFinalizer() error {
```
NautiluX: In know names tend to get lengthy, but that shouldn't be an issue with a decent IDE, so I'd prefer to not use abbreviations as they may make the code harder to understand. In this case, I even think it's obvious that we're updating a Finalizer of a ProjectReference.

pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -152,6 +154,33 @@ func (r *ReferenceAdapter) updateProjectID() error {
...
+// It returns a modification to the ProjectReference instance in the cluster for the controller to handle the reconciliation.
+func (r *ReferenceAdapter) DeletePrjRefFinalizer() error {
+	finalizers := r.projectReference.GetFinalizers()
+	if clusterapi.Contains(finalizers, prjRefFinalizer) {
+		r.projectReference.SetFinalizers(clusterapi.Filter(finalizers, prjRefFinalizer))
+		return r.kubeClient.Update(context.TODO(), r.projectReference)
+	}
+	return nil
+}
+
```
NautiluX: I really like those two functions. Maybe worth thinking if we could generalize them to use for `ProjectClaim` and `ProjectReference`?

pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -178,6 +207,34 @@ func (r *ReferenceAdapter) checkRequirements() error {
...
+	case "LIFECYCLE_STATE_UNSPECIFIED":
+		r.logger.Error(operrors.ErrUnexpectedLifecycleState, "Unexpected LifecycleState", project.LifecycleState)
+		return operrors.ErrUnexpectedLifecycleState
+	case "ACTIVE":
+		r.logger.Info("Deleting Project")
+		_, err := r.gcpClient.DeleteProject(project.ProjectId)
+		if err != nil {
+			r.logger.Error(err, "could not delete project", project.Name, "with ProjectID", project.ProjectId)
+			return err
+		}
```
NautiluX: is this logger output worth the extra if? The operator would log it anyhow. Without it's just return err :)

pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -324,6 +380,32 @@ func (r *ReferenceAdapter) createCredentials() error {
...
 }
 
+func (r *ReferenceAdapter) deleteCredentials() error {
+	secret := types.NamespacedName{
+		Namespace: r.projectClaim.Spec.GCPCredentialSecret.Namespace,
+		Name:      r.projectClaim.Spec.GCPCredentialSecret.Name,
+	}
+
+	// Check if the Secret exists or it's already deleted
+	if !gcputil.SecretExists(r.kubeClient, secret.Name, secret.Namespace) {
```
NautiluX: I'm confused, `!gcputil.SecretExists` suggests we enter this branch if the secret **does not** exist. But the content of the branch tell me you're trying to delete it there, so it should be executed when it **does** exist.

pkg/controller/projectreference/projectreference_controller.go
```diff
@@ -125,6 +125,32 @@ func (r *ReconcileProjectReference) Reconcile(request reconcile.Request) (reconc
 		return r.requeueOnErr(err)
 	}
 
+	// Check if the user (or any other mechanism) has requested to delete ProjectReference CR
```
NautiluX: obvious comments tend to get outdated soon ;)

pkg/controller/projectreference/projectreference_controller.go
```diff
@@ -125,6 +125,32 @@ func (r *ReconcileProjectReference) Reconcile(request reconcile.Request) (reconc
...
+		}
+
+		// Delete the Finalizer
+		errFinalizer := adapter.DeletePrjRefFinalizer()
+		if errFinalizer != nil {
+			return r.requeueOnErr(err)
+		}
+
+		// Cleanup is done
+		return r.doNotRequeue()
```
NautiluX: could we wrap this into a function? `EnsureProjectCleanedUp` or something?

pkg/controller/projectreference/projectreference_adapter.go
NautiluX: Also, the name suggests we're always adding a finalizer, but actually it _ensures it exists_.

pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -324,6 +380,32 @@ func (r *ReferenceAdapter) createCredentials() error {
...
+		// Get the secret key
+		key, err := gcputil.GetSecret(r.kubeClient, secret.Name, secret.Namespace)
+		if err != nil {
+			r.logger.Error(err, "could not get the service account cred secret ", "Service Account Secret Name", secret.Name)
```
NautiluX: I think we shouldn't use abbreviations in error messages

pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -152,6 +154,33 @@ func (r *ReferenceAdapter) updateProjectID() error {
...
+func (r *ReferenceAdapter) AddPrjRefFinalizer() error {
+	if !clusterapi.Contains(r.projectReference.GetFinalizers(), prjRefFinalizer) {
+		r.projectReference.SetFinalizers(append(r.projectReference.GetFinalizers(), prjRefFinalizer))
+		return r.kubeClient.Update(context.TODO(), r.projectReference)
+	}
+	return nil
+}
+
+// DeletePrjRefFinalizer parses the meta.Finalizers of ProjectReference instance and removes prjRefFinalizer if found;
+// It returns a modification to the ProjectReference instance in the cluster for the controller to handle the reconciliation.
```
NautiluX: This comment sounds like the function would return a ProjectReference, but it returns an error.

pkg/controller/projectreference/projectreference_adapter.go
drpaneas: Good idea. I am working on that.

drpaneas:
The PR is finally ready for re-review :tada:

* I have tested the functionality :sparkles: of the operator in GCP console and indeed it deletes the project. 
* Tests :white_check_mark:  have been added as well (thx to @NautiluX for the pair programming session) for both the `ReferenceController` (testing the reconciliation logic) and the `ReferenceAdapter` (testing the individual functions).

:exclamation: Regarding generalising of function that ensure the deletion/addition of the finaliser for both ProjectClaim and ProjectReference,  would prefer to deal with that in another PR.

PR #38, opened 2020-03-18
ProjectClaim controller communicates errors and updates ProjectClaimCondition

boranx:
implements: OSD-3065

### DIFF
--- a/pkg/controller/projectclaim/customresourceadapter.go
+++ b/pkg/controller/projectclaim/customresourceadapter.go
@@ -2,10 +2,12 @@ package projectclaim
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/go-logr/logr"
 	"github.com/openshift/cluster-api/pkg/util"
 	gcpv1alpha1 "github.com/openshift/gcp-project-operator/pkg/apis/gcp/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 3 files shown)


pkg/util/conditions.go
```diff
@@ -0,0 +1,56 @@
...
+// specified condition type in the given list. If none exists, then returns nil.
+func FindProjectClaimCondition(conditions []gcpv1alpha1.ProjectClaimCondition, conditionType gcpv1alpha1.ProjectClaimConditionType) *gcpv1alpha1.ProjectClaimCondition {
+	for i, condition := range conditions {
+		if condition.Type == conditionType {
+			return &conditions[i]
+		}
+	}
+
+	return nil
+}
```
NautiluX: I think those should go into the CustomResourceAdapter. You could create the adapter in the Reconcile function and pass it on to the ReconcileHandler instead of the ProjectClaim I think.

boranx: done

pkg/controller/projectclaim/projectclaim_controller.go
```diff
@@ -83,9 +80,29 @@ func (r *ReconcileProjectClaim) Reconcile(request reconcile.Request) (reconcile.
 	}
 
 	adapter := NewCustomResourceAdapter(instance, reqLogger, r.client)
+	result, err := r.ReconcileHandler(adapter)
+	if err != nil {
+		message := err.Error()
+		reason := "ReconcileFailed"
+		// Update the CRD
+		adapter.projectClaim.Status.Conditions = adapter.SetProjectClaimCondition(
```
NautiluX: I don't understand why we return the condition? Doesn't the function update the projectClaim inplace?

pkg/controller/projectclaim/customresourceadapter.go
```diff
@@ -168,3 +169,54 @@ func (c *CustomResourceAdapter) EnsureProjectClaimState(state gcpv1alpha1.ClaimS
...
+		existingCondition.Reason = reason
+		existingCondition.Message = message
+		existingCondition.LastProbeTime = now
+	}
+	return conditions
+}
+
+// FindProjectClaimCondition finds in the condition that has the
+// specified condition type in the given list. If none exists, then returns nil.
+func (c *CustomResourceAdapter) FindProjectClaimCondition() *gcpv1alpha1.ProjectClaimCondition {
```
NautiluX: Shouldn't the type be passed as an argument? At least that's how I read the comment above.

pkg/controller/projectclaim/projectclaim_controller.go
boranx: that was coming from my first POC. Fixed it.

pkg/controller/projectclaim/customresourceadapter.go
boranx: Now, we don't have to pass it. it currently uses the actual adapter's fields.

NautiluX: Now I don't understand the comment anymore :D

boranx: well, me too :D

pkg/controller/projectclaim/projectclaim_controller.go
```diff
@@ -83,9 +80,25 @@ func (r *ReconcileProjectClaim) Reconcile(request reconcile.Request) (reconcile.
 	}
 
 	adapter := NewCustomResourceAdapter(instance, reqLogger, r.client)
+	result, err := r.ReconcileHandler(adapter)
+	if err != nil {
+		message := err.Error()
+		reason := "ReconcileFailed"
+		// Update the ProjectClaimConditionCRD
+		adapter.SetProjectClaimCondition(corev1.ConditionTrue, reason, message)
```
rafael-azevedo: We need to update the status of the ProjectClaim after updating the conditions so that we write the conditions to the CR.

boranx: the status update is now being made in `SetProjectClaimCondition()`

pkg/controller/projectclaim/customresourceadapter.go
```diff
@@ -161,10 +163,65 @@ func (c *CustomResourceAdapter) EnsureProjectClaimState(state gcpv1alpha1.ClaimS
...
 	c.projectClaim.Status.State = state
+	return c.StatusUpdate()
+}
+
+// StatusUpdate updates the project claim status
+func (c *CustomResourceAdapter) StatusUpdate() error {
 	err := c.client.Status().Update(context.TODO(), c.projectClaim)
 	if err != nil {
-		c.logger.Error(err, "Failed to update ProjectClaim state")
+		c.logger.Error(err, fmt.Sprintf("failed status update for %s", c.projectClaim.Name))
```
NautiluX: +1 adding the name of the ProjectClaim to the error message. However, this new error message lacks the information that we tried to upgrade a ProjectClaim state, I think that can be helpful information.

pkg/controller/projectclaim/projectclaim_controller.go
```diff
@@ -83,9 +80,25 @@ func (r *ReconcileProjectClaim) Reconcile(request reconcile.Request) (reconcile.
 	}
 
 	adapter := NewCustomResourceAdapter(instance, reqLogger, r.client)
+	result, err := r.ReconcileHandler(adapter)
+	if err != nil {
+		message := err.Error()
+		reason := "ReconcileFailed"
+		// Update the ProjectClaimConditionCRD
+		return result, adapter.SetProjectClaimCondition(corev1.ConditionTrue, reason, message)
```
boranx: should we return the actual err instead of StatusUpdate() err?

NautiluX: Definitely! I think errors occuring here, we should just ignore.

boranx: anyway, just ignored

PR #42, opened 2020-03-23
Fix ProjectReference CR yaml

drpaneas:
The name of the ProjectReference CR is not corresponding to the  name generated by the ProjectClaim CR. This PR applies the same logic as gcp-project-operator would have done it:

  name: `<namespace of projectclaim cr>-<name of projectclaim cr>`

See [here](https://github.com/openshift/gcp-project-operator/blob/master/pkg/controller/projectclaim/customresourceadapter.go#L40)


drpaneas:
indeed, it's more confusing having it hardcoded as an example like that.

PR #44, opened 2020-03-24
Add finalizer before GCP project setup

drpaneas:
* Add the ProjectReference's finalizer before the GCP project
      configuration.
  * Add requeueAfter() method to ProjectClaim controller
  * ProjectClaim is not deleted as long as ProjectReference exists.
  * Modify customresourceadapter's tests with the new logic

### DIFF
--- a/pkg/controller/projectclaim/customresourceadapter.go
+++ b/pkg/controller/projectclaim/customresourceadapter.go
@@ -69,24 +69,44 @@ func (c *CustomResourceAdapter) IsProjectClaimDeletion() bool {
 	return c.projectClaim.DeletionTimestamp != nil
 }
 
-func (c *CustomResourceAdapter) FinalizeProjectClaim() error {
-	projectReferenceExists, err := c.ProjectReferenceExists()
-	if err != nil {
-		return err
-	}
+func (c *CustomResourceAdapter) IsProjectReferenceDeletion() bool {
+	return c.projectReference.DeletionTimestamp != nil
+}
 
-	if projectReferenceExists {
-		err := c.client.Delete(context.TODO(), c.projectReference)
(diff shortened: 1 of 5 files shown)


drpaneas:
I think I've figured a way to handle this. Right now, when you issue a delete request for ProjectClaim, the finalizer is basically dumb: [it deletes itself if it exists](https://github.com/openshift/gcp-project-operator/blob/master/pkg/controller/projectclaim/customresourceadapter.go#L82-L85).

I have fixed that locally in my setup, so now if you issue a ProjectClaim deletion, the claim's adapter checks if there is a projectereference it doesn't allow the finalizer to get deleted. Hence, nothing gets deleted.

The next step which I am still working, is that we need to teach ProjectReference to delete the finalizer of ProjectClaim instance after its EnsureCleanUp function has no errors. In that way, the ProjectReference will be deleted only when (the token, its finalizer, and the project in gcp are deleted).

tldr: To create a project in GCP you create a claim cr. To shutdown that project, you need to delete the project-reference (which will delete the claim cr and not the other way around).

^ does this make sense? If so, I can start implementing it.

NautiluX:
I think we're just missing a `requeueAfter()` in the ProjectClaim finalization [here|https://github.com/openshift/gcp-project-operator/blob/9f2a90b81dde61853a55c489358b7fb2974dfc34/pkg/controller/projectclaim/customresourceadapter.go#L80] make it wait for the ProjectReference to get deleted.
I guess the [`DELETE` call](https://github.com/openshift/gcp-project-operator/blob/9f2a90b81dde61853a55c489358b7fb2974dfc34/pkg/controller/projectclaim/customresourceadapter.go#L77) may return without the resource actually being removed.

PR #46, opened 2020-04-01
Add availability zone to project claim

rafael-azevedo:
This PR adds a list of availability zones to the ProjectClaim CRD and adds the methods to do so in the client and project reference controller.


rafael-azevedo:
Take a look at the last commit. It contains all the logic. The previous commits contain vendored and auto generated files to make reviewing easier.

pkg/gcpclient/client.go
```diff
@@ -97,17 +101,46 @@ func NewClient(projectName string, authJSON []byte) (Client, error) {
...
+	err := req.Pages(context.Background(), func(page *compute.ZoneList) error {
+		for _, zone := range page.Items {
+			if strings.Contains(zone.Region, region) {
+				zones[region] = append(zones[region], zone.Name)
+			}
+		}
+		return nil
+	})
+	if err != nil {
+		fmt.Println("ahhhhh")
```
NautiluX, line 136: best error message on the operator so far.

pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -469,6 +480,26 @@ func (r *ReferenceAdapter) deleteCredentials() error {
 	return nil
 }
 
+func (r *ReferenceAdapter) checkAvailibilityZones() error {
```
NautiluX, line 483: I think this function should be called ensureAvailabilityZonesSet() or sometthing similar. Also I think it should return if it modified the object or not. Read comment above for why.

pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -117,16 +118,25 @@ func (r *ReferenceAdapter) EnsureProjectClaimUpdated() (gcpv1alpha1.ClaimStatus,
...
+	if len(r.projectClaim.Spec.AvailibilityZones) <= 0 {
+		err := r.checkAvailibilityZones()
 		if err != nil {
-			r.logger.Error(err, "Error updating ProjectClaim GCPProjectID")
+			r.logger.Error(err, "Error updating chceking availibility zones")
 			return r.projectClaim.Status.State, err
 		}
 	}
 
+	err := r.kubeClient.Update(context.TODO(), r.projectClaim)
```
NautiluX, line 131: I think we should only call update if the `ProjectClaim` actually changed. Therefore, the functions called should inform the caller about that.

pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -117,16 +118,25 @@ func (r *ReferenceAdapter) EnsureProjectClaimUpdated() (gcpv1alpha1.ClaimStatus,
 
 	if r.projectClaim.Spec.GCPProjectID == "" {
 		r.projectClaim.Spec.GCPProjectID = r.projectReference.Spec.GCPProjectID
-		err := r.kubeClient.Update(context.TODO(), r.projectClaim)
+	}
+
+	if len(r.projectClaim.Spec.AvailibilityZones) <= 0 {
```
NautiluX, line 123: this is checked in the function called as well, which I think is better.

pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -117,16 +118,25 @@ func (r *ReferenceAdapter) EnsureProjectClaimUpdated() (gcpv1alpha1.ClaimStatus,
 
 	if r.projectClaim.Spec.GCPProjectID == "" {
```
NautiluX, line 119: we could put this into its own function with similar signature than the function that adds availability zones, so we have a single level of abstraction in this function:
```
if  !ready:
  return

modified = ensureClaimProjectIdSet()
modified = modified || ensureClaimAvailabilityZonesSet()
if modified:
  updateClaim()
...
```

pkg/gcpclient/client.go
```diff
@@ -97,17 +101,44 @@ func NewClient(projectName string, authJSON []byte) (Client, error) {
...
 		serviceManagmentClient:     serviceManagmentClient,
 		cloudBillingClient:         cloudBillingClient,
+		computeClient:              computeService,
 		credentials:                creds,
 	}, nil
 }
 
+// ListAvilibilityZones returns a map of all avilibility zones a project has access to
+// where the key is the region and the values is a list of zones
+func (c *gcpClient) ListAvilibilityZones(projectID, region string) (map[string][]string, error) {
```
NautiluX, line 123: If I understand this function correctly it only lists the zones of the given regions. So, given input region "x", with matching zones "z1" and "z2" it will return
```
{
"x": ["z1", "z2"]
}
```
why not just return an array `["z1", "z2"]`?

rafael-azevedo: The reason I put this into a map was to support additional regions/zones in the future , but if its not necessary now I can remove it.

PR #49, opened 2020-04-03
Remove ClusterDeployment controller

ninabauer:


pkg/util/mocks/structs/builders.go
```diff
@@ -22,9 +22,8 @@ const (
 	testProject = "project"
 	testRegion  = "us-east1"
 
-	clusterPlatformLabel          = "hive.openshift.io/cluster-platform"
-	clusterPlatformGCP            = "gcp"
-	clusterDeploymentManagedLabel = "api.openshift.com/managed"
+	clusterPlatformLabel = "hive.openshift.io/cluster-platform"
+	clusterPlatformGCP   = "gcp"
```
rafael-azevedo, line 26: These two variables can be removed.

pkg/util/util.go
```diff
@@ -38,25 +38,6 @@ func GetSecret(kubeClient client.Client, secretName, namespace string) (*corev1.
...
-			Namespace: namespace,
-		},
-		Data: map[string][]byte{
-			"osServiceAccount.json": []byte(creds),
-		},
-	}
-}
-
 // NewGCPSecretCRV2 returns a Secret CR formatted for GCP for use in projectreference controller.
 func NewGCPSecretCRV2(creds string, namespacedNamed kubetypes.NamespacedName) *corev1.Secret {
```
rafael-azevedo, line 42: Can we rename this function to  NewGCPSecretCR and adjust the comments as well

pkg/util/mocks/structs/builders.go
rafael-azevedo: Actually I think this whole file can be removed

pkg/util/util.go
ninabauer: Renamed function and comment

pkg/util/mocks/structs/builders.go
ninabauer: Removed the variables.

PR #50, opened 2020-04-06
use hive.openshift.io/v1

staebler:


NautiluX:
Hi @staebler, thanks for your PR! I checked the dependency to hive and it seems we can remove that entirely. So I think with #51 we can close this PR if you agree?

staebler:
Even better. Thanks.

/close

PR #52, opened 2020-04-08
projectreference controller communicates errors

boranx:
Implements: OSD-3067

This PR provides common condition struct for both `projectclaim` and `projectreference`.
The conditions will be marked as resolved if  `ReconcileHandler()` returns no err for the next run.

Example handle flow:

```shell
oc apply -f example-claim.yaml
watch oc get projectreference --all-namespaces -o yaml | grep -A10  conditions
```
The expected output should be like the following:
```shell
conditions: []
    state: Creating
.
.
.
    conditions:
    - lastProbeTime: "2020-04-27T13:14:39Z"
      lastTransitionTime: "2020-04-27T13:14:37Z"
      message: 'googleapi: Error 403: Access Not Configured. Compute Engine API has
        not been used in project XXXXXXX before or it is disabled. Enable it
        by visiting https://console.developers.google.com/apis/api/compute.googleapis.com/overview?project=XXXXXXX
        then retry. If you enabled this API recently, wait a few minutes for the action
        to propagate to our systems and retry., accessNotConfigured'
      reason: ReconcileError
      status: "True"
      type: Error
    state: Ready
.
.
.
    conditions:
    - lastProbeTime: "2020-04-27T13:14:48Z"
      lastTransitionTime: "2020-04-27T13:14:37Z"
      message: 'googleapi: Error 403: Access Not Configured. Compute Engine API has
        not been used in project XXXXXXX before or it is disabled. Enable it
        by visiting https://console.developers.google.com/apis/api/compute.googleapis.com/overview?project=XXXXXXX
        then retry. If you enabled this API recently, wait a few minutes for the action
        to propagate to our systems and retry., accessNotConfigured'
      reason: ReconcileError
      status: "True"
      type: Error
    state: Ready
.
.
.
    conditions:
    - lastProbeTime: "2020-04-27T13:15:52Z"
      lastTransitionTime: "2020-04-27T13:14:37Z"
      message: 'googleapi: Error 403: Access Not Configured. Compute Engine API has
        not been used in project XXXXXXX before or it is disabled. Enable it
        by visiting https://console.developers.google.com/apis/api/compute.googleapis.com/overview?project=XXXXXXX
        then retry. If you enabled this API recently, wait a few minutes for the action
        to propagate to our systems and retry., accessNotConfigured'
      reason: ReconcileErrorResolved
      status: "False"
      type: Error
    state: Ready
```


pkg/controller/projectreference/projectreference_controller.go
```diff
@@ -137,14 +149,14 @@ func (r *ReconcileProjectReference) Reconcile(request reconcile.Request) (reconc
...
+	if adapter.projectReference.Status.State == "" {
 		reqLogger.Info("Checking Requirements")
 		err := adapter.checkRequirements()
 		if err != nil {
 			// TODO: add condition here SupportedRegion = false to give more information on the error state
 			reqLogger.Error(err, "Region not supported")
-			projectReference.Status.State = gcpv1alpha1.ProjectReferenceStatusError
-			err := r.client.Status().Update(context.TODO(), projectReference)
+			adapter.projectReference.Status.State = gcpv1alpha1.ProjectReferenceStatusError
+			err := r.client.Status().Update(context.TODO(), adapter.projectReference)
```
NautiluX, line 159: We would need to initialize the ProjectReferenceCondition here as well

pkg/controller/projectreference/projectreference_controller.go
```diff
@@ -154,15 +166,16 @@ func (r *ReconcileProjectReference) Reconcile(request reconcile.Request) (reconc
 
 		reqLogger.Info(fmt.Sprintf("Setting ProjectReferenceStatus %s", gcpv1alpha1.ProjectReferenceStatusCreating))
 		// passed requirementes check set to creating
-		projectReference.Status.State = gcpv1alpha1.ProjectReferenceStatusCreating
-		err = r.client.Status().Update(context.TODO(), projectReference)
+		adapter.projectReference.Status.State = gcpv1alpha1.ProjectReferenceStatusCreating
+		adapter.projectReference.Status.Conditions = []gcpv1alpha1.ProjectReferenceCondition{}
```
NautiluX, line 170: Maybe do it similar to the ProjectClaim, where we ensure the condition is initalized once at the beginning? https://github.com/openshift/gcp-project-operator/blob/6e30cdb0d2cf9e9fa03f40e2a09858f6a47f2c5f/pkg/controller/projectclaim/projectclaimadapter.go#L112

pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -574,6 +576,63 @@ func (r *ReferenceAdapter) SetIAMPolicy(serviceAccountEmail string) error {
...
+			existingConditions.LastTransitionTime = now
+		}
+		existingConditions.Status = status
+		existingConditions.Reason = reason
+		existingConditions.Message = message
+		existingConditions.LastProbeTime = now
+	}
+
+	return r.statusUpdate()
+}
```
NautiluX, line 610: I like that we're doing this similar to the ProjectReference handling, however, this is a siginificant amount of code duplication (https://github.com/openshift/gcp-project-operator/blob/6e30cdb0d2cf9e9fa03f40e2a09858f6a47f2c5f/pkg/controller/projectclaim/projectclaimadapter.go#L201), I think it would be nice to have this unified, if possible. Otherwise we risk the two functions evolve independently, which might become a hotbed for bugs..

boranx: @NautiluX that function's logic is exported to `util.go` and callable from either projectreference or projectclaim. Can you have a look at that again?

pkg/apis/gcp/v1alpha1/common_types.go
```diff
@@ -11,3 +16,37 @@ type NamespacedName struct {
 	Name      string `json:"name"`
 	Namespace string `json:"namespace"`
 }
+
+// Condition contains details for the current condition of a custom resource
+type Condition struct {
```
NautiluX, line 21: I like the unification of the Condition type.

pkg/util/util.go
```diff
@@ -154,3 +155,50 @@ func InArray(needle interface{}, haystack interface{}) (exists bool, index int)
 
 	return
 }
+
+// SetCondition sets a condition on a custom resource's status
+func SetCondition(conditions *[]gcpv1alpha1.Condition, status corev1.ConditionStatus, reason string, message string) error {
```
NautiluX, line 160: Would be nice to have tests for this function, as it's part of the public interface of util.go.

boranx: added one for that

pkg/controller/projectreference/projectreference_adapter_test.go
```diff
@@ -125,6 +126,49 @@ var _ = Describe("ProjectreferenceAdapter", func() {
...
+							if condition.Type == gcpv1alpha1.ConditionError {
+								found = &conditions[i]
+							}
+						}
+
+						Expect(message).To(Equal(found.Message))
+						Expect(reason).To(Equal(found.Reason))
+
+						//Hold the last state
+						firstLastProbeTime = found.LastProbeTime
```
boranx, line 157: I can remove that since the logic has been moved to `util.go`

pkg/util/util.go
NautiluX: is one test good enough for this function? Did you cover all the different routes? Cyclomatic complexity is 4, so I would think the number of tests should be somewhere around that?

boranx: well, I've tested the main functionality here, not all edge cases. For me that test covers the ability to change the `conditions` array properly. What do you mean by _all the different routes_?

NautiluX: If you say your test covers everything, I'm fine :)

pkg/util/util.go
```diff
@@ -14,6 +14,67 @@ import (
...
+		existingCondition.LastProbeTime = now
+	}
+
+	return nil
+}
+
+// findProjectClaimCondition finds the suitable Condition object
+// by looking for adapter's condition list.
+// If none exists, then returns nil.
+func (u *gcpUtil) findProjectClaimCondition(conditions []gcpv1alpha1.Condition) *gcpv1alpha1.Condition {
```
NautiluX, line 67: this is not related to ProjectClaim anymore, is it? If not, it should become a new name.

NautiluX: Also I think the method could be public

pkg/util/util.go
```diff
@@ -14,6 +14,67 @@ import (
...
+type gcpUtil struct {
+}
+
+// NewUtil returns a valid util object with bound functions that can be mockable
+func NewUtil() Util {
+	return &gcpUtil{}
+}
+
+// SetCondition sets a condition on a custom resource's status
+func (u *gcpUtil) SetCondition(conditions *[]gcpv1alpha1.Condition, status corev1.ConditionStatus, reason string, message string) error {
```
NautiluX, line 32: I like that you change this to be a method of a struct instead of a utility function. However, I think it shouldn't be called Util but related to the Condition type? `ConditionManager` or something like that? If the State object was unified, I'd say make it a method of that type. But it's not, and probably better won't be.

pkg/util/util.go
boranx: done

pkg/util/util.go
boranx: it's now moved to `condition/conditions.go`

boranx:
@drpaneas thanks, I guess this can be considered as initial an we can improve the usage of condition but for now it's ready to go

boranx:
@drpaneas are you sure that you are on the last commit? 
The state should be either `status: IssueIsOngoing` or `status: Resolved`

drpaneas:
> @drpaneas are you sure that you are on the last commit?
> The status should be either `status: IssueIsOngoing` or `status: Resolved`

Indeed. I was testing a previous commit. Now it seems fixed:

```
status:
  conditions:
  - lastProbeTime: "2020-04-24T09:59:58Z"
    lastTransitionTime: "2020-04-24T09:58:14Z"
    message: 'googleapi: Error 403: Access Not Configured. Compute Engine API has
      not been used in project 719402616192 before or it is disabled. Enable it by
      visiting https://console.developers.google.com/apis/api/compute.googleapis.com/overview?project=719402616192
      then retry. If you enabled this API recently, wait a few minutes for the action
      to propagate to our systems and retry., accessNotConfigured'
    reason: ProjectReferenceReconcileHandlerFailed
    status: Resolved
    type: Error
  state: Ready
```

pkg/condition/conditions.go
```diff
@@ -0,0 +1,61 @@
...
+			},
+		)
+	} else {
+		if status == corev1.ConditionTrue {
+			existingCondition.Status = "IssueIsOngoing"
+			existingCondition.Reason = reason
+			existingCondition.Message = message
+			existingCondition.LastProbeTime = now
+		} else {
+			existingCondition.Status = "Resolved"
```
NautiluX, line 45: the type of the status field is corev1.ConditionStatus, I think we shouldn't assign a string to it.

NautiluX: Can we use True and False again here(true: ongoing, false: resolved), and use the Reason field to make it visible when the error resolved? (`ReconcileError` and `ReconcileErrorResolved` probably?)

NautiluX: That would also make the `SetCondition` function more generic, currently it would be only useful for the error condition (which is the only one we have so far).

pkg/condition/conditions.go
```diff
@@ -0,0 +1,62 @@
...
+				Type:               conditionType,
+				Status:             status,
+				Reason:             reason,
+				Message:            message,
+				LastTransitionTime: now,
+				LastProbeTime:      now,
+			},
+		)
+	} else {
+		if status == corev1.ConditionTrue {
```
NautiluX, line 39: I think you can remove this whole if now

pkg/condition/conditions.go
boranx: @NautiluX done, now we're using `reason` field for that.

pkg/condition/conditions.go
boranx: I've made some changes here, can you have a look @NautiluX ?

PR #55, opened 2020-04-15
s/availibility/availability/

mmazur:
If this doesn't get fixed, we'll get an endless stream of irritated SREs that get irritated when they notice this.

On the other hand I don't know what are the real world implications of merging it this late. Hence do-not-merge/hold.

### DIFF
--- a/deploy/crds/gcp_v1alpha1_projectclaim_crd.yaml
+++ b/deploy/crds/gcp_v1alpha1_projectclaim_crd.yaml
@@ -42,7 +42,7 @@ spec:
           type: object
         spec:
           properties:
-            availibilityZones:
+            availabilityZones:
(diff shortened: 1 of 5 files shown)


2uasimojo:
Is this thing in prod yet? If not, we should definitely fix it. Letting this kind of typo escape to the field is bad for many reasons (not just SRE irritation).

/lgtm

2uasimojo:
/lgtm cancel

I infer from @NautiluX's comment that this has already escaped to the field :(

NautiluX:
Yes it has escaped. We have a backlog item to fix it (OSD-3422).

NautiluX:
closing, as the 3-stage-solution is being worked on.

PR #56, opened 2020-04-16
Enable regions with sufficient quota

NautiluX:


pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -67,19 +67,20 @@ var supportedRegions = map[string]bool{
...
 
-	// The regions below are all currently
-	// They do not have enough quota configured by default
-	// "asia-east2":              true,
-	// "asia-northeast2":         true,
-	// "asia-south1":             true,
-	// "australia-southeast1":    true,
-	// "europe-north1":           true,
-	// "europe-west2":            true,
+	// Regions below don't have enough quota configured by default, but our org has sufficient quota
```
drpaneas, line 70: Do you think we should document what "enough quota" means for future reference? Apart from that lgtm

NautiluX: think you're right, will create a follow-up PR

PR #58, opened 2020-04-16
Change ProjectReference adapter test package

NautiluX:
* Make API public
* Add tests for previously private functions

### DIFF
--- a/pkg/controller/projectreference/projectreference_adapter.go
+++ b/pkg/controller/projectreference/projectreference_adapter.go
@@ -25,7 +25,7 @@ import (
 
 const (
 	osdServiceAccountName = "osd-managed-admin"
-	finalizerName         = "finalizer.gcp.managed.openshift.io"
+	FinalizerName         = "finalizer.gcp.managed.openshift.io"
 )
 
 // OSDRequiredAPIS is list of API's, required to setup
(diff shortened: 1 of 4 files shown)


pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -85,61 +85,61 @@ var supportedRegions = map[string]bool{
...
-	projectReference *gcpv1alpha1.ProjectReference
+	ProjectClaim     *gcpv1alpha1.ProjectClaim
+	ProjectReference *gcpv1alpha1.ProjectReference
 	logger           logr.Logger
 	kubeClient       client.Client
 	gcpClient        gcpclient.Client
 }
 
-func newReferenceAdapter(projectReference *gcpv1alpha1.ProjectReference, logger logr.Logger, client client.Client, gcpClient gcpclient.Client) (*ReferenceAdapter, error) {
+func NewReferenceAdapter(projectReference *gcpv1alpha1.ProjectReference, logger logr.Logger, client client.Client, gcpClient gcpclient.Client) (*ReferenceAdapter, error) {
```
drpaneas, line 95: Needs a comment to above the function `// NewReferenceAdapater ....`

pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -206,37 +206,37 @@ func getMatchingClaimLink(projectReference *gcpv1alpha1.ProjectReference, client
 }
 
 // updateProjectID updates the ProjectReference with a unique ID for the ProjectID
-func (r *ReferenceAdapter) updateProjectID() error {
+func (r *ReferenceAdapter) UpdateProjectID() error {
```
drpaneas, line 209: Fix the comment `// UpdateProjectID`

speaking of it, it's not much trouble fix the comment also for https://github.com/openshift/gcp-project-operator/blob/master/pkg/controller/projectreference/projectreference_adapter.go#L276 which is referring to the wrong function.

pkg/controller/projectreference/projectreference_adapter.go
NautiluX: I dislike this so much.

pkg/controller/projectreference/projectreference_adapter.go
NautiluX: we don't need a comment for a private function, do we?

PR #65, opened 2020-05-07
add resource requests/limits per OSD-2109

bmeng:
current resource usage:
gcp-project-operator-756d6f99dc-l9wmx   4m           71Mi

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -26,7 +26,7 @@ spec:
               memory: "100Mi"
             limits:
               cpu: "50m"
-              memroy: "200Mi"
+              memory: "200Mi"
           env:
             - name: WATCH_NAMESPACE
               value: ""
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -20,6 +20,13 @@ spec:
           command:
           - gcp-project-operator
           imagePullPolicy: Always
+          resources:
+            requests:
+              cpu: "10m"
+              memory: "100Mi"
+            limits:
+              cpu: "50m"
+              memroy: "200Mi"
           env:
             - name: WATCH_NAMESPACE
               value: ""
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -22,11 +22,11 @@ spec:
           imagePullPolicy: Always
           resources:
             requests:
-              cpu: "10m"
-              memory: "100Mi"
+              cpu: "500m"
+              memory: "500Mi"
             limits:
-              cpu: "50m"
-              memory: "200Mi"
+              cpu: "500m"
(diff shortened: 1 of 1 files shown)


deploy/operator.yaml
```diff
@@ -20,6 +20,13 @@ spec:
           command:
           - gcp-project-operator
           imagePullPolicy: Always
+          resources:
+            requests:
+              cpu: "10m"
+              memory: "100Mi"
+            limits:
+              cpu: "50m"
+              memroy: "200Mi"
```
drpaneas, line 29:
```suggestion
              memory: "200Mi"
```

bmeng:
updated, thanks!

drewandersonnz:
@bmeng can you explain why you've chosen request/limit ratio of 5x for CPU and 2x for RAM? 

This is something our customer have been doing and it has caused immense pain for them. 

I think we should specify 1.0x ratio (requests and limits match), unless there is a very good reason to do this differently.

bmeng:
> @bmeng can you explain why you've chosen request/limit ratio of 5x for CPU and 2x for RAM?
> 
> This is something our customers have been doing and it has caused immense pain for them.
> 
> I think we should specify 1.0x ratio (requests and limits match), unless there is a very good reason to do this differently.

I do not have a good reason for this, basically I'd like to keep both cpu and memory have the same ratio 2x to make sure the service will not be in problem in peak.
But as the current usage of the cpu is too low, I am afraid that the 2x for limit of request will not be safe.

@drewandersonnz What is your suggestion of setting both the requests and limits here?

deploy/operator.yaml
```diff
@@ -20,6 +20,13 @@ spec:
           command:
           - gcp-project-operator
           imagePullPolicy: Always
+          resources:
+            requests:
+              cpu: "10m"
```
NautiluX, line 25:
```suggestion
              cpu: "500m"
```

deploy/operator.yaml
```diff
@@ -20,6 +20,13 @@ spec:
...
+              memory: "100Mi"
```
NautiluX, line 26:
```suggestion
              memory: "500Mi"
```

deploy/operator.yaml
```diff
@@ -20,6 +20,13 @@ spec:
...
+            limits:
+              cpu: "50m"
```
NautiluX, line 28:
```suggestion
              cpu: "500m"
```

deploy/operator.yaml
```diff
@@ -20,6 +20,13 @@ spec:
...
+              memory: "200Mi"
```
NautiluX, line 29:
```suggestion
              memory: "500Mi"
```

PR #66, opened 2020-05-19
move contributing to docs/, add an example pr template

boranx:


georgettica:
so I found this to validate the commit https://github.com/conventional-changelog/commitlint

we might want to add automatic linting with husky, found this blog that explains how to use hooks with the tool https://www.vojtechruzicka.com/githooks-husky/

hack/app_sre_pr_check.sh
```diff
@@ -11,6 +11,9 @@ if [ "$?" != "0" ]; then
     exit 1
 fi
 
+# Check PR Commit message for being conformant
+docker run --volume $(pwd):/app --rm gtramontina/commitlint:8.3.5 --from=HEAD~1
```
NautiluX, line 15: I'm not sure if I'd like to see the build script fail on a commit message format, having to look up the reason for the failure on Jenkins.
If we want to have something like this, I'd recommend adding it as commit hook to catch it earlier and easier to identify. But this will make it harder to do quick and dirty commits in local branches (need to add `--no-verify`).

Also, this will only check the last commit message, not the all of a PR's commits if I'm right?

boranx: @NautiluX Yes first of all, we are still not sure adding that into the PR's context. Wondering your thoughts... I recommend instead of adding local hooks to git, install commitlint to your local and execute it against your commit message. That can be added to the contributing docs but folks can easily find the problem by following the pipeline logs. 
Eventually, they will be redirected to https://github.com/conventional-changelog/commitlint/#what-is-commitlint (which contains the only one rule that needs to be followed for now).
The format is generic as well: https://www.conventionalcommits.org/en/v1.0.0/
The thing we need to do is tagging the commit with a suitable item.

For question two, considering we're squashing the commits into a single one before the merge, it should be enough for the pr. However, we can still apply the rules for all commit

NautiluX: I'm pretty sure many PRs will include more than 1 commit, so yes we should cover all the commits

docs/CONTRIBUTING.md
```diff
@@ -25,3 +25,38 @@ If you are new to open-source contribution, please read the this [guide](https:/
...
+- [ ] the latest changes against crc/k8s
+- [ ] the `make coverage` command to generate new calculated coverage
+```
+
+## PR Review Checks
+
+This section isn't mandatory for all PR's but it would be nice to check the followings after reviewing the source code as a post-check.
+
+- checkout the code and run the operator against your own crc/k8s, make sure the provided feature is working as expected
+- make sure that the coverage badge will be updated in the PR's context
```
NautiluX, line 60: how?

docs/CONTRIBUTING.md
```diff
@@ -25,3 +25,38 @@ If you are new to open-source contribution, please read the this [guide](https:/
...
+- run `make` and make sure there are no code changes(especially in `go.mod`, `go.sum` and for the other generated files)
```
NautiluX, line 61: I think `make` will fail if there are dependencies that are not in the `vendor` folder. Also, make is executed by the CI, so no need to do that as reviewer

docs/CONTRIBUTING.md
```diff
@@ -25,3 +25,38 @@ If you are new to open-source contribution, please read the this [guide](https:/
...
+- run `go mod vendor` and make sure the additional dependencies are located in `vendor/` folder.
```
NautiluX, line 62: See comment above

docs/CONTRIBUTING.md
```diff
@@ -25,3 +25,38 @@ If you are new to open-source contribution, please read the this [guide](https:/
...
+Fixes #
+
+Special notes for your reviewer:
+
+Is it a breaking change or backward compatible?
+
+Pre-checks:
+I've run
+- [ ] the `make` command and it returns 0
+- [ ] the latest changes against crc/k8s
```
NautiluX, line 51:
```suggestion
- [ ] manually tested latest changes against crc/k8s
```

docs/CONTRIBUTING.md
```diff
@@ -25,3 +25,38 @@ If you are new to open-source contribution, please read the this [guide](https:/
...
+
+Fixes #
+
+Special notes for your reviewer:
+
+Is it a breaking change or backward compatible?
+
+Pre-checks:
+I've run
+- [ ] the `make` command and it returns 0
```
NautiluX, line 50: will be done by CI

PR #71, opened 2020-05-26
Wait for compute API to get ready without error

NautiluX:
* Don't error if compute API is not ready when querying availability
    zones
* Set condition telling compute API is not ready
* If the compute API is not ready after 10 minutes, return error
* Enable mock tear-down checks in projectreference_adapter_test


pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -467,28 +473,55 @@ func (r *ReferenceAdapter) deleteCredentials() error {
...
 
 	r.ProjectClaim.Spec.AvailibilityZones = zones
 
-	return true, nil
+	return EnsureAzResultModified, nil
+}
+
+func (r *ReferenceAdapter) handleAvailabilityZonesError(err error) (EnsureAzResult, error) {
+	conditions := &r.ProjectReference.Status.Conditions
+	if strings.HasPrefix(err.Error(), "googleapi: Error 403: Access Not Configured. Compute Engine API has not been used in project") {
```
drpaneas, line 505: I hope google doesn't change this error message.

pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -467,28 +473,55 @@ func (r *ReferenceAdapter) deleteCredentials() error {
...
+		return EnsureAzResultNoChange, nil
 	}
 
 	zones, err := r.gcpClient.ListAvilibilityZones(r.ProjectReference.Spec.GCPProjectID, r.ProjectClaim.Spec.Region)
 	if err != nil {
-		return false, err
+		return r.handleAvailabilityZonesError(err)
 	}
+	conditions := &r.ProjectReference.Status.Conditions
+	r.conditionManager.SetCondition(conditions, gcpv1alpha1.ConditionComputeApiReady, corev1.ConditionTrue, "QueryAvailabilityZonesSucceeded", "ComputeAPI ready, successfully queried availability zones")
```
georgettica, line 496: is there a global place to find constants is this repo? I searched but it seems the state `QueryAvailabilityZonesSucceeded` is only saved where the code is called, I expected to have a consts file with all of them

pkg/controller/projectreference/projectreference_adapter.go
NautiluX: unfortunately we don't have a type or anything to recognize the error, other than this message...

pkg/controller/projectreference/projectreference_adapter.go
NautiluX: as long as we don't need it in another package, I don't see much reason to move it. What do you think would be the benefit of moving it?

georgettica: to have all of the statuses in one place so we can identify where each state is used

pkg/controller/projectreference/projectreference_adapter.go
georgettica: maybe add a comment that explains it so if it changes we know it's fragile

georgettica:
```
// This error message is not backed in any real error type, so working with string comparison is our best approach
```

NautiluX: not sure a comment like this describes it better than the actual code

pkg/controller/projectreference/projectreference_adapter.go
NautiluX: As long as this is not used outside the module, I prefer to keep it in this file. That being said, I think it should be private.

pkg/controller/projectreference/projectreference_adapter.go
NautiluX: tried to make that clear with a function

PR #78, opened 2020-06-08
[WIP] feat(monitoring): add monitoring binary

georgettica:
inside the ./monitoring folder we have monitoring for the amount of ProjectClaim

### What type of PR is this? 
feature

Add a custom monitoring to the `gcp-project-operator`

OSD-3147

currently, to run this you will need to use it external to the cluster, use `go run monitoring/monitoring.go` to start the server.

### Pre-checks:


go.mod
```diff
@@ -5,7 +5,7 @@ require (
 	contrib.go.opencensus.io/exporter/ocagent v0.4.9 // indirect
 	github.com/Azure/go-autorest v11.5.2+incompatible // indirect
 	github.com/appscode/jsonpatch v0.0.0-20190108182946-7c0e3b262f30 // indirect
-	github.com/coreos/prometheus-operator v0.26.0 // indirect
+	github.com/coreos/prometheus-operator v0.0.0-00010101000000-000000000000 // indirect
```
RiRa12621, line 8: Are you sure about the version?

build/Dockerfile
```diff
@@ -1,14 +1,18 @@
...
-ENV OPERATOR_PATH=/gcp-project-operator \
-    OPERATOR_BIN=gcp-project-operator
+ENV OPERATOR_BIN=gcp-project-operator \
+    OPERATOR_PATH=/${OPERATOR_BIN}
+ARG  OPERATOR_BASE_PATH=/gcp-project-operator/build/_output/bin
 
 WORKDIR /root/
-COPY --from=builder /gcp-project-operator/build/_output/bin/${OPERATOR_BIN} /usr/local/bin/${OPERATOR_BIN}
+COPY --from=builder ${OPERATOR_BASE_PATH}/gcp-project-operator /usr/local/bin/gcp-project-operator
+COPY --from=builder ${OPERATOR_BASE_PATH}/monitoring /usr/local/bin/monitoring
```
RiRa12621, line 15: I'd personally prefer the "one process per container" approach over putting the operator and a separate monitoring binary to run in the same container.

monitoring/localmetrics/localmetrics.go
```diff
@@ -0,0 +1,64 @@
...
+func NewMetricsConfig(c client.Client, log logr.Logger) *MetricsConfig {
+	return &MetricsConfig{c: c, log: log}
+}
+
+func (m MetricsConfig) PublishMetrics() {
+	m.TotalProjectClaims()
+	go func() {
+		for {
+			select {
+			case <-time.After(3 * time.Second):
```
RiRa12621, line 46: If there's only one case, could this rather be an `if` for simplicity?

monitoring/localmetrics/localmetrics.go
```diff
@@ -0,0 +1,64 @@
...
+				err := m.TotalProjectClaims()
+				if err != nil {
+					m.log.Error(err, "Cannot Expose metrics to prometheus")
```
RiRa12621, line 49: I wouldn't mention "Prometheus" here as it may draw wrong conclusions during debugging. What fails is the exposing, regardless if or if not anything related to prometheus works

go.mod
georgettica: nope, didn't change it directly. I will try to revert and see if it works

build/Dockerfile
georgettica: This was an option but this is the best way to tie the version of the operator to the monitoring binary.
I can also do the separation  but that involves creating another quay repo.

monitoring/localmetrics/localmetrics.go
georgettica: it can and I will do that :)

monitoring/localmetrics/localmetrics.go
georgettica: good point, I will remove the mention to prometheus

go.mod
georgettica: works, bumped to latest

monitoring/localmetrics/localmetrics.go
georgettica: what I did instead is add the stop channel, I just forgot to add it here

monitoring/localmetrics/localmetrics.go
georgettica: done

georgettica:
After I get /lgtm or /approve I will squash all of these commits

build/Dockerfile
NautiluX: Having it in the same container image doesn't mean the container necessarily runs both. If it is a significant overhead to have a separate image, I'd say we can do it like this. However, I think we need to add another `container` to the operator [deployment](https://github.com/openshift/gcp-project-operator/blob/master/deploy/operator.yaml) for that.

PR #80, opened 2020-06-18
Refactor Reconcile

NautiluX:
### What type of PR is this? 
refactor

* Use defined function definition and return value for all operations performed on ProjectReference
* Operations can control behavior of the operator
* Do not throw errors if GCP service account is not initialized
* Reduce number of parallel reconciles by returning after updating a CR
* Reduce number of API calls to GCP by querying for a list of enabled APIs
* Fix repeatedly creating service account keys when compute API is not ready
* When secret is created, we can assume service API key is created. No need to create a new key and secret.

_Fixes #OSD-3889_

### Pre-checks:
- manually tested latest changes against crc/k8s


boranx:
I liked the way of handling with `StopProcessing()` and `ContinueProcessing()`
/lgtm

pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -387,37 +445,52 @@ func (r *ReferenceAdapter) configureServiceAccount() error {
...
 	}
 
-	return nil
+	return ContinueProcessing()
 }
 
-func (r *ReferenceAdapter) createCredentials() error {
-	var serviceAccount *iam.ServiceAccount
+func matchesAlreadyExistsError(err error) bool {
+	return strings.HasPrefix(err.Error(), "googleapi: Error 409:")
```
georgettica, line 467: group `match*Error` to a single place

pkg/controller/projectreference/projectreference_controller.go
```diff
@@ -121,67 +122,58 @@ func (r *ReconcileProjectReference) Reconcile(request reconcile.Request) (reconc
...
+	CancelRequest  bool
+}
 
-	// Make projectReference  be processed based on state of ProjectClaim and Project Reference
-	claimStatus, err := adapter.EnsureProjectClaimReady()
-	if claimStatus == gcpv1alpha1.ClaimStatusReady || err != nil {
-		return r.requeueOnErr(err)
-	}
+func StopProcessing() (OperationResult, error) {
+	return OperationResult{0 * time.Second, false, true}, nil
```
georgettica, line 133: does this still work?
```suggestion
	return OperationResult{0, false, true}, nil
```

georgettica, line 133: what about this?
```suggestion
	return OperationResult{0 * time.Second, RequeueRequest: false, CancelRequest : true}, nil
```

pkg/controller/projectreference/projectreference_controller_test.go
```diff
@@ -211,6 +211,8 @@ var _ = Describe("ProjectReference controller reconcilation", func() {
 		Context("When Reference State is empty and it failes to update", func() {
```
georgettica, line 211:
```suggestion
		Context("When Reference State is empty and it fails to update", func() {
```

pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -101,46 +102,81 @@ func NewReferenceAdapter(projectReference *gcpv1alpha1.ProjectReference, logger
...
-			return r.ProjectClaim.Status.State, operrors.Wrap(err, "error updating ProjectClaim spec")
+			return RequeueWithError(operrors.Wrap(err, "error updating ProjectClaim spec"))
 		}
+		return StopProcessing()
 	}
 
 	if azResult == ensureAzResultNotReady {
-		return r.ProjectClaim.Status.State, nil
+		r.logger.V(int(logtypes.ProjectReference)).Info("Compute API not yet fully initialized. Retrying in 30 seconds.")
+		return RequeueAfter(30*time.Second, nil)
```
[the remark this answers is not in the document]
NautiluX: gofmt doesn't seem to allow that space

pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -101,46 +102,81 @@ func NewReferenceAdapter(projectReference *gcpv1alpha1.ProjectReference, logger
...
-func (r *ReferenceAdapter) EnsureProjectConfigured() error {
+func VerifyProjectClaimPending(r *ReferenceAdapter) (OperationResult, error) {
+	if r.ProjectClaim.Status.State != gcpv1alpha1.ClaimStatusPendingProject {
+		return RequeueAfter(5*time.Second, nil)
+	}
+	return ContinueProcessing()
+}
+
+func EnsureProjectReferenceStatusCreating(adapter *ReferenceAdapter) (OperationResult, error) {
+	if adapter.ProjectReference.Status.State == "" {
```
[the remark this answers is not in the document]
NautiluX: ack

pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -101,46 +102,81 @@ func NewReferenceAdapter(projectReference *gcpv1alpha1.ProjectReference, logger
...
+			err = operrors.Wrap(err, "error updating ProjectReference status")
+			return RequeueWithError(err)
+		}
+		return StopProcessing()
+	}
+	return ContinueProcessing()
+}
+
+func EnsureProjectID(adapter *ReferenceAdapter) (OperationResult, error) {
+	if adapter.ProjectReference.Spec.GCPProjectID == "" {
```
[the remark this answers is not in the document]
NautiluX: yep

pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -149,40 +185,41 @@ func (r *ReferenceAdapter) EnsureProjectConfigured() error {
...
 		r.logger.V(int(logtypes.ProjectReference)).Info("Setting Status on projectReference")
 		r.ProjectReference.Status.State = gcpv1alpha1.ProjectReferenceStatusReady
-		return r.kubeClient.Status().Update(context.TODO(), r.ProjectReference)
+		return RequeueOnError(r.kubeClient.Status().Update(context.TODO(), r.ProjectReference))
 	}
-	return nil
+	return ContinueProcessing()
 }
 
 func getMatchingClaimLink(projectReference *gcpv1alpha1.ProjectReference, client client.Client) (*gcpv1alpha1.ProjectClaim, error) {
```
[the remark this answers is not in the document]
NautiluX: I think we will reuse it for ProjectClaim controller, that means we'll put it somewhere central

pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -356,9 +412,11 @@ func (r *ReferenceAdapter) configureAPIS(config configmap.OperatorConfigMap) err
 	}
 
 	for _, a := range OSDRequiredAPIS {
```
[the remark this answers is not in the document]
NautiluX: renamed it to api as loop variable, renamed apis to enabledAPIs

pkg/controller/projectreference/projectreference_adapter.go
NautiluX: agree

pkg/controller/projectreference/projectreference_controller.go
NautiluX: it does! Thanks

NautiluX: As the struct is declared 10 lines above, I think this is not necessary as long as it stays with 3 parameters.

pkg/controller/projectreference/projectreference_controller_test.go
NautiluX: thx

pkg/controller/projectreference/projectreference_controller.go
georgettica: I think that adding these inside the util function allows people to peek inside the fucntion and discern what boolean mean what without going to the  struct decleration

pkg/controller/projectreference/projectreference_adapter.go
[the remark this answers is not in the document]
NautiluX: I don't like the additional line this would add everywhere. I think it's obvious from the function call with or without space. I'd do RequeueWithLongDelay(nil). But I also like that we actually show the time we delay.

pkg/controller/projectreference/projectreference_controller.go
NautiluX: Okay, I don't like the verboseness this creates, but I agree it might be easier to read. Will name the parameters.

pkg/controller/projectreference/projectreference_adapter.go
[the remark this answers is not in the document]
georgettica: agreed, let's leave it at that.
maybe extract to constants?
`shortDelay` and `longDelay`?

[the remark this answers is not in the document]
georgettica: nvm. let's just leave at that and worry about it in another day

drpaneas:
I came late to the party, but I didn't have any objections. Great work :)

PR #84, opened 2020-06-25
Refactor ProjectClaim controller

NautiluX:
### What type of PR is this? 

Refactoring

* unify subroutines called in Reconcile
* match pattern from ProjectClaim reconcilation
* move utility functions and result to util package

### Pre-checks:

### DIFF
--- a/pkg/controller/projectclaim/projectclaim_controller.go
+++ b/pkg/controller/projectclaim/projectclaim_controller.go
@@ -6,6 +6,8 @@ import (
 
 	gcpv1alpha1 "github.com/openshift/gcp-project-operator/pkg/apis/gcp/v1alpha1"
 	condition "github.com/openshift/gcp-project-operator/pkg/condition"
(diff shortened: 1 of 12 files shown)


pkg/controller/projectclaim/projectclaim_controller.go
```diff
@@ -101,55 +104,33 @@ func (r *ReconcileProjectClaim) Reconcile(request reconcile.Request) (reconcile.
 	return result, err
 }
 
+type ReconcileOperation func() (util.OperationResult, error)
```
drpaneas, line 107: I don't think I understand that. Can you explain a bit what purpose it fits?
Should this be `util.OperationResult, error` or `gcputil.OperationResult, error`

NautiluX: that's confusing. Here we import that package as util.

boranx:
I've tested it, that looks good.

PR #87, opened 2020-07-09
Changing base image for Dockerfile

ravitri:
### What type of PR is this? 
_bug_

PR is raised to use build base image from Redhat which would be supported and to change the operator base image to ubi8.

_Complies to #_ https://issues.redhat.com/browse/OSD-2383

Set Golang version to 1.13 for now and not 1.14. Can change if required.

Backward compatible


georgettica:
I think it looks fine.. what is required for a local build?

PR #88, opened 2020-07-09
fix(log_level): reorganized log levels

georgettica:
now log levels are sorted by verbosity, the higher the number the more verbose it is

### What type of PR is this? 
fix
this clarifies the logs are sorted by level

`--zap-level=1` == `--zap-level=debug`

and each log with a higher number needs to be set by
`--zap-level=<NUMBER>`

_Fixes [OSD-4381](https://issues.redhat.com/browse/OSD-4381)_

it is backwards comaptible and not not breaking 
### Pre-checks:


pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -468,22 +466,22 @@ func (r *ReferenceAdapter) createCredentials() (gcputil.OperationResult, error)
 		return gcputil.ContinueProcessing()
 	}
 
+	r.logger.Info("Creating credentials")
 	serviceAccount, err := r.gcpClient.GetServiceAccount(osdServiceAccountName)
 	if err != nil {
 		if matchesNotFoundError(err) {
-			r.logger.V(int(logtypes.ProjectReference)).Info("Service Account not yet fully initialized. Retrying in 30 seconds.")
+			r.logger.V(2).Info("Service Account not yet fully initialized. Retrying in 30 seconds.")
```
NautiluX, line 473:
```suggestion
			r.logger.V(1).Info("Service Account not yet fully initialized. Retrying in 30 seconds.")
```

georgettica: without V() is V(0) which means display in prod

georgettica:
> Why do some log messages have `V()` and some not? What will the default level be if you just call logger.Info?

without `V()` it is `V(0)`

PR #89, opened 2020-07-27
Update run command

drpaneas:
- Flag --local has been deprecated, use 'run local' instead
 - Flag --namespace is deprecated; use --watch-namespace instead.

### What type of PR is this? 
_docs_

because we updated to the new operator-sdk version.

### Pre-checks:


georgettica:
this is version dependant. maybe put a note that this is in the newer versions

drpaneas:
> this is version dependant. maybe put a note that this is in the newer versions

done, thx

georgettica:
once more :)

PR #91, opened 2020-08-11
OSD-4792 Fix Catalog img build script

rogbas:
### What type of PR is this? 
bug

Updates Catalog Img build script to reference the new app-interface repo, instead of the old saas-osd-operators when getting the latest prod git commit
OSD-4792


rogbas:
cc @NautiluX

PR #102, opened 2020-09-09
Boilerplate bootstrap

2uasimojo:
### What type of PR is this? 
refactor

Initial setup of [boilerplate](https://github.com/openshift/boilerplate/) with the [openshift/golang-osd-operator](https://github.com/openshift/boilerplate/tree/aae6e37ad3eb05bd5855693b11560858e81c8e9d/boilerplate/openshift/golang-osd-operator) convention.

Things of note:
- project.mk and standard.mk are wholly replaced by their counterparts in the boilerplate convention.
- The `gocheck` target brought in by boilerplate uses golangci-lint instead of gofmt + go vet. This introduced some failures, which are addressed here. Some of the fixes are legit; some are spoofed, with TODOs added, to be addressed later; and some (in tests) are simply bypassed.
- The `coverage` target is replaced by the [standard codecov script/config/target](https://github.com/openshift/ops-sop/blob/9e8fbd5c6233118de1825bde65eb766bb943089a/services/codecov.md). Note that the prow and codecov.io infrastructure for this flavor of code coverage is not yet set up for this repo at this time. (See [OSD-5135](https://issues.redhat.com/browse/OSD-5135).)
- The previous gopherbadger-based coverage target is renamed `gopherbadger` for now. Once this repo is set up with codecov.io, that can be removed, and we can simply link to the badge hosted at codecov.io. (Ditto.)
- The `generate` target used to be just for generating go stuff. Now that's done via the `gogenerate` target. It's also done generically, and doesn't explicitly load up the dependencies it used to. As a result, running the old target modifies go.mod and go.sum (but nothing else) but running the new `gogenerate` target is currently a no-op.
- The boilerplate `generate` target does both `gogenerate` as described above, and operator-sdk stuff. As it stands, that results in nontrivial changes, which appear to be the result of having upgraded operator-sdk versions without regenerating (at least) CRDs. We're not messing with that here. [Later] Actually it's from me having a more recent version of operator-sdk installed locally. Boilerplate is still figuring out how to standardize on that.

_Fixes #_ [OSD-4943](https://issues.redhat.com/browse/OSD-4943)

Should be 100% backward compatible. The actual code changes are trivial. And by the time this is ready to merge, any CI problems will be fixed.

### Pre-checks:
- [x] run the `make gopherbadger` command to generate new calculated coverage

I didn't do the first thing, since this is just a framework change.

Note renaming in the second thing for now. Eventually that checkbox will go away entirely.


NautiluX (commented):
/approve
Side note: I still do not understand why we can't use submodules for this, it would mean a lot less changes and additional files in each repo, as this PR already clearly shows.

RiRa12621:
/hold
I don't see why we should approve this, if we're not clear on the reasoning of not using submodules @NautiluX

2uasimojo:
I believe it's on @jharrington22's (long) to-do list to produce a more detailed explanation for why TΩ+leads decided early on to avoid submodules for boilerplate. But my understanding of the main reason is that it requires developers of the consuming repositories to understand certain things about submodules and add certain flags to "normal" git commands even when they're not explicitly doing work related to the submodules. And we deemed it not reasonable to expect to educate everyone on those arcana, as well as making sure they have the right versions of git installed.

RiRa12621:
/unhold
Fair enough for now, gonna wait for the detailed explanation.
However I would have preferred if decisions of that impact to everyone would be made in the open

2uasimojo:
Superseded by #124.

PR #103, opened 2020-09-10
Add support for CCS console access for groups

yeya24:
Signed-off-by: Ben Ye <yb532204897@gmail.com>

### What type of PR is this? 
feature & refactory

1. As now both CCS and non-CCS needs to get the operator configmap, I move this step to the beginning of the reconcile and keep the config in the adapter struct.

2. Refactor the configmap format to yaml in order to better support a list of groups that need to get ccs console access. Now the configmap format becomes like

``` yaml
apiVersion: v1
data:
  data: |
    billingAccount: xxx
    parentFolderID: xxx
    ccsConsoleAccess:
    - example-group@xxx.com
kind: ConfigMap
metadata:
  name: gcp-project-operator
  namespace: gcp-project-operator
```

3. Unit tests updated

Fixes https://issues.redhat.com/browse/OSD-5041 https://issues.redhat.com/browse/OSD-5046

### Pre-checks:


README.md
```diff
@@ -135,16 +135,21 @@ Note: unless you're running this against your very own GCP org, **someone likely
...
+kind: ConfigMap
+metadata:
+  name: gcp-project-operator
+  namespace: gcp-project-operator
+data:
+  data: |
+    billingAccount: "123456-ABCDEF-123456" # Google billing ID from https://console.cloud.google.com/billing
+    parentFolderID: "123456789123"         # Google Cloud organization Parent Folder ID
+    ccsConsoleAccess:
+    - sd-sre-platform-gcp-access@redhat.com # A list of groups that you want to give CCS console access to
```
mwoodson, line 153: We don't want the actual group name anywhere.  Let's put an example in here.

docs/gcpconfig.md
```diff
@@ -13,12 +13,22 @@ If you don't have one, please [create](https://cloud.google.com/billing/docs/how
...
+kind: ConfigMap
+metadata:
+  name: gcp-project-operator
+  namespace: gcp-project-operator
+data:
+  data: |
+    billingAccount: "123456-ABCDEF-123456" # Google billing ID from https://console.cloud.google.com/billing
+    parentFolderID: "123456789123"         # Google Cloud organization Parent Folder ID
+    ccsConsoleAccess:
+    - sd-sre-platform-gcp-access@redhat.com # A list of groups that you want to give CCS console access to
```
mwoodson, line 30: again, let's remove the actual group name.

pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -67,6 +67,19 @@ var OSDRequiredRoles = []string{
 	"roles/compute.admin",
 }
 
+// CCSConsoleAccessRequiredRoles is a list of Roles that a service account
+// required to get console access of CCS Accounts
+var CCSConsoleAccessRequiredRoles = []string{
```
mwoodson, line 72: please see https://docs.google.com/document/d/1vqZeayNt-GUsLJslw9jTmORvUPf1WvdhRcZgZJQeCIU/edit# for the latest list of roles

README.md
yeya24: Sure, I will update it.

pkg/configmap/configmap_test.go
```diff
@@ -30,18 +31,34 @@ func TestValidateOperatorConfigMap(t *testing.T) {
...
+					},
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "gcp-project-operator",
+						Namespace: "gcp-project-operator",
+					},
+					Data: map[string]string{
+						"data": `
+parentFolderID: 1234567
+billingAccount: billing123
+`,
```
mwoodson, line 59: dumb question, is this formatted properly?

pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -67,6 +67,20 @@ var OSDRequiredRoles = []string{
 	"roles/compute.admin",
 }
 
+// OSDSREConsoleAccessRoles is a list of Roles that a service account
+// required to get console access.
+var OSDSREConsoleAccessRoles = []string{
+	"roles/compute.admin",
+	//"roles/iam.organizationRoleAdmin",
+	"roles/editor",
+	//"roles/resourcemanager.organizationViewer",
```
mwoodson, line 76: If these aren't needed, we should delete them.

pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -226,6 +247,17 @@ func EnsureProjectConfigured(r *ReferenceAdapter) (gcputil.OperationResult, erro
 		return gcputil.RequeueWithError(operrors.Wrap(err, "error creating credentials"))
 	}
 
+	if r.isCCS() {
+		r.logger.V(1).Info("Configuring Service Account Permissions for Console Access")
```
mwoodson, line 251: Is this message correct?  Is the operator configuring the service account?  Or is it configuring IAM to allow console access?

pkg/configmap/configmap_test.go
yeya24: Yes

pkg/controller/projectreference/projectreference_adapter.go
yeya24: They are needed, but seems `roles/resourcemanager.organizationViewer` is deprecated and `roles/iam.organizationRoleAdmin` cannot be created using the `SetIAM` function. @sam-nguyen7

pkg/controller/projectreference/projectreference_adapter.go
yeya24: `Configuring IAM to allow console access` is better! I will update it.

mwoodson:
This PR looks good to me.  Would want @rafael-azevedo  and/or @NautiluX to also look

README.md
```diff
@@ -135,16 +135,21 @@ Note: unless you're running this against your very own GCP org, **someone likely
 ### Configmap
 
 The controller expects to find a `ConfigMap` with the name `gcp-project-operator` inside the `gcp-project-operator` namespace.
-It will parse it and verify its contents, expecting to extract the values of two specific fields that should be already populated by you:
 
-* `billingAccount`
-* `parentFolderID`
+You can easily create one example `ConfigMap` using the following command.
```
NautiluX, line 139:
```suggestion
You can create one example `ConfigMap` using the following command.
```

README.md
```diff
@@ -135,16 +135,21 @@ Note: unless you're running this against your very own GCP org, **someone likely
...
+kind: ConfigMap
+metadata:
+  name: gcp-project-operator
+  namespace: gcp-project-operator
+data:
+  data: |
+    billingAccount: "123456-ABCDEF-123456" # Google billing ID from https://console.cloud.google.com/billing
+    parentFolderID: "123456789123"         # Google Cloud organization Parent Folder ID
+    ccsConsoleAccess:
+    - example-group@xxx.com # A list of groups that you want to give CCS console access to
```
NautiluX, line 153:
```suggestion
    - example-group@xxx.com # A list of groups that will get access to CCS projects
```

docs/gcpconfig.md
```diff
@@ -13,12 +13,22 @@ If you don't have one, please [create](https://cloud.google.com/billing/docs/how
 For parent folder you can use any folder you like.
 If you don't have one, feel free to [create](https://cloud.google.com/resource-manager/docs/creating-managing-folders) one.
 
-Add this information to your Kubernetes cluster by creating a `ConfigMap`:
+You can easily create one example `ConfigMap` using the following command.
```
NautiluX, line 16:
```suggestion
You can create one example `ConfigMap` using the following command.
```

docs/gcpconfig.md
```diff
@@ -13,12 +13,22 @@ If you don't have one, please [create](https://cloud.google.com/billing/docs/how
...
+kind: ConfigMap
+metadata:
+  name: gcp-project-operator
+  namespace: gcp-project-operator
+data:
+  data: |
+    billingAccount: "123456-ABCDEF-123456" # Google billing ID from https://console.cloud.google.com/billing
+    parentFolderID: "123456789123"         # Google Cloud organization Parent Folder ID
+    ccsConsoleAccess:
+    - example-group@xxx.com # A list of groups that you want to give CCS console access to
```
NautiluX, line 30:
```suggestion
    - example-group@xxx.com # A list of groups that will get access to CCS projects
```

NautiluX: Also, since this is copied over, can't we reference it from the README or the other way round?

pkg/configmap/configmap.go
```diff
@@ -42,12 +41,16 @@ func ValidateOperatorConfigMap(configmap OperatorConfigMap) error {
 func GetOperatorConfigMap(kubeClient client.Client) (OperatorConfigMap, error) {
 	var OperatorConfigMap OperatorConfigMap
```
NautiluX, line 42: btw, why does this variable name start with a capital o?

pkg/configmap/configmap.go
```diff
@@ -42,12 +41,16 @@ func ValidateOperatorConfigMap(configmap OperatorConfigMap) error {
...
 	configmap := &corev1.ConfigMap{}
-	if err := kubeClient.Get(context.TODO(), kubetypes.NamespacedName{Name: OperatorConfigMapName, Namespace: operatorConfigMapNamespace}, configmap); err != nil {
+	if err := kubeClient.Get(context.TODO(), kubetypes.NamespacedName{Name: OperatorConfigMapName, Namespace: OperatorConfigMapNamespace}, configmap); err != nil {
 		return OperatorConfigMap, fmt.Errorf("unable to get configmap: %v", err)
 	}
 
-	if err := mapstructure.Decode(configmap.Data, &OperatorConfigMap); err != nil {
-		return OperatorConfigMap, fmt.Errorf("unable to unmarshal configmap: %v", err)
+	if data, ok := configmap.Data[OperatorConfigMapKey]; !ok {
```
NautiluX, line 48: this will try to get `configmap.Data["data"]` right? Is this correct?

pkg/configmap/configmap_test.go
NautiluX: Agree it's difficult to read, maybe use a one-line json?

pkg/configmap/configmap_test.go
```diff
@@ -82,6 +129,41 @@ func TestGetOperatorConfigMap(t *testing.T) {
...
+						Namespace: "gcp-project-operator",
+					},
+					Data: map[string]string{
+						"data": `
+parentFolderID: 1234567
+billingAccount: billing123
+ccsConsoleAccess:
+- foo
+- bar
+`,
```
NautiluX, line 151: see above. Using json may increase readability

pkg/controller/projectreference/projectreference_adapter.go
NautiluX: that means this code is still WIP? Or you want to merge it like this for now but we know it won't work for future changes?

pkg/controller/projectreference/projectreference_adapter.go
```diff
@@ -635,13 +654,13 @@ func (r *ReferenceAdapter) AddOrUpdateBindings(serviceAccountEmail string) (Addo
 }
 
 // SetIAMPolicy attempts to update policy if the policy needs to be modified
-func (r *ReferenceAdapter) SetIAMPolicy(serviceAccountEmail string) error {
+func (r *ReferenceAdapter) SetIAMPolicy(serviceAccountEmail string, policies []string, group bool) error {
```
NautiluX, line 657: I'd prefer to not use a bool as type for group, this makes function calls hard to read. I recommend using a type for it.

NautiluX (changes requested):
Hi @yeya24, thanks for submitting this PR! I made a few comments. My biggest concern is the format of the config map, currently it looks like
```
apiVersion: v1
data:
  billingAccount: asdf
  orgParentFolderID: asdf
  parentFolderID: asdf
kind: ConfigMap
metadata:
  annotations:
  name: gcp-project-operator
  namespace: gcp-project-operator
```

Are you sure the new structs are able to parse this kind of config map? Because in the example and implementation it looks like there was an additional level in them, which confuses me:

```
data:
  data: |
```

pkg/controller/projectreference/projectreference_adapter.go
yeya24: This part is blocked https://coreos.slack.com/archives/CMD1L32M9/p1600780878019500. I think it is okay to ignore this first. WDYT? I can add a TODO.

pkg/configmap/configmap.go
yeya24: Nice catch!

pkg/configmap/configmap.go
yeya24: Yes, it requires the `data` key.

README.md
yeya24: Thanks! It is much better!

docs/gcpconfig.md
yeya24: Great idea. I will add a link from readme to the config doc.

pkg/controller/projectreference/projectreference_adapter.go
sam-nguyen7: I will be reaching out to the google TAM to follow up on this. @yeya24 you can remove these roles for this PR. Thank you!

pkg/controller/projectreference/projectreference_adapter.go
yeya24: WDYT @rafael-azevedo @sam-nguyen7 ?

pkg/configmap/configmap_test.go
yeya24: Umm, why this is a problem. That's how `` is used to express yaml. It would be clear in the real yaml config.

yeya24:
> Are you sure the new structs are able to parse this kind of config map? Because in the example and implementation it looks like there was an additional level in them, which confuses me:

It cannot parse the old format and the config map needs to be updated.

pkg/configmap/configmap.go
NautiluX: Can't we find a way to parse the existing format? I like the parsing, but having
```
data:
  data:
```
seems unclean to me.

pkg/controller/projectreference/projectreference_adapter.go
NautiluX: would it be a high effort to get this done here? I just fear code we add with the intention to change "later" because often enough later means never unfortunately.

yeya24: Makes sense! But tbh I cannot figure out that API and it is not a high priority. So I will change bool to a type first.

pkg/configmap/configmap.go
yeya24: It is hard to still parse the existing format. Since config map itself is `map[string]string`, if we want to have a list, then we only have 2 choices:

1. Have a single key, and put all our configs to the value. If you think the key name `data` not good, I can update it to something else, like `config.yaml`, so people will know it is actually a yaml config file. WDYT?

``` yaml
data:
  config.yaml: |
    billingAccount: xxx
    folderID: xxx
    consoleAccess: [item1, item2]
```

2. Have multiple keys. In this case, we have to find a way to make consoleAccess value a string. TBH it might doable, but maybe mess up things.

``` yaml
data:
  billingAccount: xxx
  folderID: xxx
  consoleAccess: "item1, item2"
```

pkg/controller/projectreference/projectreference_adapter.go
yeya24: Updated, sorry for the delay. Please take a look!

pkg/configmap/configmap.go
yeya24: I am still looking at this and I find a way to get this working for both formats.

``` yaml
data:
  billingAccount: xxx
  folderID: xxx
  config.yaml: |
    billingAccount: xxx
    folderID: xxx
    consoleAccess: [item1, item2]
```

yeya24:
Can I get another review for this pr? @NautiluX @sam-nguyen7

pkg/configmap/configmap.go
NautiluX: yes @yeya24 let's do it like this and switch to the new format now, I think it's the cleaner way of defining and parsing a configmap. Some thoughts:
- Make sure it works if only the "old" fields are set
- Don't call it `data`, maybe `config.yaml` would be a good alternative.
- Let's not defer cleaning up the redundant fields but prepare a PR right away so we can clean it up as soon as all environments are migrated.

lisa: 👍 to `config.yaml` key name.

NautiluX (approved):
/lgtm
/approve
/hold pending deployment of new config maps

NautiluX:
/unhold
config maps updated

PR #114, opened 2020-10-07
Remove vendor/

2uasimojo:
Get rid of in-repo vendored dependencies


georgettica:
I think that would be good, I'd this is the way all the other repos are intended to be

RiRa12621:
Let's quickly look at what's happening here:
there's a lot of clutter when changes are being made and that's not really great I agree. However mod=vendor saves us of anything going wrong in the dependencies in any way shape or form.
When dependencies go missing like recently happened on https://github.com/openshift/operator-custom-metrics we are on the safe side of things for example.
Therefore I don't think this change should go in.

2uasimojo:
@NautiluX [agrees](https://coreos.slack.com/archives/C0189RRTQAV/p1602103304129300?thread_ts=1602095555.122000&cid=C0189RRTQAV):
> Vendoring dependencies makes code reviews a bit more challenging, but github abstracts that away very well. On the other side, vendoring speeds up builds as you don't need to go get everything on every build step but you have it bundled with check in. Also your build doesn't fail if an external dependency is not available. So, overall I'm in favor of vendoring dependencies so the code is self contained. I'm not sure if we need to make this decision in common for all repos we maintain tbh.

PR #124, opened 2020-11-23
Boilerplate bootstrap at d9d5abfd0745651959b5e4de82b751ec4c3a3dac

2uasimojo:
This PR comprises multiple commits:
- Bootstrap boilerplate (generated commit)
- Accommodate `make` target changes
- Make `generate` checks work
- Make `lint` checks work

See the individual commit messages for details.


2uasimojo (commented):
/hold

Needs a fixup commit.

2uasimojo:
> /hold
> 
> Needs a fixup commit.

This comment was left by mistake

/hold cancel

jharrington22:
Hey @NautiluX what is the status here with this PR? We are working towards onbaording all the operator operators to use the OLM builds within boilerplate, this is a requirement to support image digests so we need to do the same here. How can we help?

NautiluX:
I rebased and updated boilerplate, @2uasimojo for when you're online, can you please check if it all makes sense? It's a 7.5k+ change so I'm afraid I could miss something. Feel free to unhold when good.
/lgtm
/hold

jharrington22:
@NautiluX Thanks, the changes look good to me I'll wait for @2uasimojo in case there was anything else he wanted to check. The makefile has a single edit and we're not overriding any of your make targets so this should be safe.

2uasimojo:
I'm just making sure the appsre pipeline is still explicitly calling the hack script. (Otherwise this update will trigger the boilerplated pipeline which will need a cleanup commit.)

2uasimojo:
https://gitlab.cee.redhat.com/service/app-interface/-/blob/86f2a8d12ec8014e221a0db317192e8e42ee656b/data/services/osd-operators/cicd/ci-int/jobs-gcp-project-operator.yaml#L27 ✓

There's a bunch of cleanup that still needs to happen here, like getting rid of project.mk and standard.mk and the Makefile targets that are now overriding what's in boilerplate, but let's go ahead and land this and we can do that stuff in a followon.

/hold cancel

PR #128, opened 2020-12-02
Remove orpolicy.policyViewer role

sam-nguyen7:
### What type of PR is this? 
cleanup
Removes the roles/orgpolicy.policyViewer from SRE console access group

_Fixes #_

### Pre-checks:


mwoodson:
I support this. We can't find a good reason to have this role, and why it is needed.  

Since this role is at the org level, not the project level, it brings more concerns to customers.  I think we should remove it here and as a requirement for GCP CCS

georgettica:
I guess the best way to test it will be to deploy to stage  && see if it's fine there..
If we are more concerned, we can do in on a recommend hive shard and force a deploy on it to see it suceeds

NautiluX (commented):
/approve
this role is applied to the SRE group, so it doesn't impact the functionality of the operator itself.

PR #131, opened 2021-01-26
[WIP] OSD-5405 - allow deletion of projectClaim without deleting the gcp project

sulmer-redhat:
This is WIP PR, do not merge.
This PR is a POC - we want to be able to let the operator know to not delete the gcp project in case where a flag is set (onThemove: true)

### What type of PR is this? 

_Fixes #_

### Pre-checks:


deploy/crds/gcp.managed.openshift.io_projectclaims_crd.yaml
```diff
@@ -1,147 +1,161 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
ravitri, line 1: Was curious about setting v1 CRD. AFAIK, we are to default to v1beta1 for time being till we have v3 hives present. Also, need to check with boilerplate team too if SRE operators can use v1 CRD now or not. 

cc: @jharrington22 @2uasimojo

georgettica: Agreed, maybe there is an option to dual version the crds?

PR #142, opened 2021-04-15
Hive operators reference a different key in SAAS files

jharrington22:
### What type of PR is this? 
bug

Current deploys reference a deployed has in the saas files. This PR fixes the location of the hash lookup.

Fixes the hash location in #141 but keeps the hash check

### Pre-checks:


NautiluX (commented):
/lgtm

PR #157, opened 2021-06-30
OSD-7174 - Add fake project claim

Tessg22:
### What type of PR is this? 
feature

As SDA developer, I want to be able to create a fake project claim that moves into ready instantly, so that I can test integration with GCP Project Operator

_Fixes #_

### Pre-checks:

### DIFF
--- a/pkg/controller/projectclaim/projectclaimadapter.go
+++ b/pkg/controller/projectclaim/projectclaimadapter.go
@@ -176,30 +176,29 @@ func (c *ProjectClaimAdapter) IsProjectClaimFake() (gcputil.OperationResult, err
 		if _, err := c.EnsureFinalizer(); err != nil {
 			return gcputil.RequeueWithError(operrors.Wrap(err, fmt.Sprintf("Could not update project claim state for %s", c.projectClaim.Name)))
 		}
-
 		// If project claim is marked for deletion, remove fake secret
-		if c.projectClaim.DeletionTimestamp != nil && gcputil.SecretExists(c.client, c.projectClaim.Spec.GCPCredentialSecret.Name, c.projectClaim.Spec.GCPCredentialSecret.Namespace) {
+		if c.projectClaim.DeletionTimestamp != nil {
 			secret := &corev1.Secret{}
-			c.client.Get(context.TODO(), types.NamespacedName{
+			err := c.client.Get(context.TODO(), types.NamespacedName{
 				Name:      c.projectClaim.Spec.GCPCredentialSecret.Name,
 				Namespace: c.projectClaim.Spec.GCPCredentialSecret.Namespace},
 				secret,
 			)
-			err := c.client.Delete(context.TODO(), secret)
+			if err != nil {
+				return gcputil.RequeueWithError(operrors.Wrap(err, fmt.Sprintf("Could not get fake secret %s", c.projectClaim.Name)))
+			}
 
-			// Confirm that secret was deleted
-			if err != nil || gcputil.SecretExists(c.client, c.projectClaim.Spec.GCPCredentialSecret.Name, c.projectClaim.Spec.GCPCredentialSecret.Namespace) {
+			err = c.client.Delete(context.TODO(), secret)
+			if err != nil {
 				return gcputil.RequeueWithError(operrors.Wrap(err, fmt.Sprintf("Could not delete fake secret %s", c.projectClaim.Spec.GCPCredentialSecret.Name)))
 			}
 
-			// Remove finalizer to unlock deletion of the project claim
-			err = c.EnsureProjectClaimFinalizerDeleted()
+			_, err = c.EnsureProjectClaimDeletionProcessed()
 			if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/controller/projectclaim/projectclaimadapter.go
```diff
@@ -41,6 +42,7 @@ const (
 const ProjectClaimFinalizer string = "finalizer.gcp.managed.openshift.io"
 const CCSSecretFinalizer string = "finalizer.gcp.managed.openshift.io/ccs"
 const RegionCheckFailed string = "RegionCheckFailed"
+const FakeProjectClaim string = "managed.openshift.com/fake"
```
ramonbutter, line 45: These fake Projects are going to be used for hive performance testing. 
Who is setting this annotation for fake ProjectClaims that are generated for fake ClusterDeployments? Is it a hive?
How did we come up with this naming? 
Shouldn't it be the same as here? : 
https://github.com/openshift/hive/blob/023f095eaf99afeb730b103f6881ef081c7ac88e/pkg/constants/constants.go#L321

pkg/controller/projectclaim/projectclaimadapter.go
```diff
@@ -169,7 +171,80 @@ func (c *ProjectClaimAdapter) FinalizeProjectClaim() (ObjectState, error) {
...
+		if !gcputil.SecretExists(c.client, c.projectClaim.Spec.GCPCredentialSecret.Name, c.projectClaim.Spec.GCPCredentialSecret.Namespace) {
+			privateKeyString, err := base64.StdEncoding.DecodeString("SS1hbS1mYWtlLXBhc3M=")
+			if err != nil {
+				return gcputil.RequeueWithError(operrors.Wrap(err, "Could not decode hardcoded secret"))
+			}
+			if err := c.client.Create(context.TODO(), gcputil.NewGCPSecretCR(string(privateKeyString), types.NamespacedName{Namespace: c.projectClaim.Spec.GCPCredentialSecret.Namespace, Name: c.projectClaim.Spec.GCPCredentialSecret.Name})); err != nil {
+				return gcputil.RequeueWithError(operrors.Wrap(err, fmt.Sprintf("Could not create fake secret %s", c.projectClaim.Spec.GCPCredentialSecret.Name)))
+			}
+		}
+		// Update fake project claim specs
```
ramonbutter, line 212: Why do we need to change the spec of the claim? Is the annotation not enough?

pkg/controller/projectclaim/projectclaimadapter.go
Tessg22: Confirmed with Manuel that `managed.openshift.com/fake` is correct.

pkg/controller/projectclaim/projectclaimadapter.go
```diff
@@ -169,7 +171,114 @@ func (c *ProjectClaimAdapter) FinalizeProjectClaim() (ObjectState, error) {
...
+		err := c.client.Status().Update(context.TODO(), c.projectClaim)
+		if err != nil {
+			return true, err
+		}
+		return false, nil
+	}
+	return true, nil
+}
+
+func (c *ProjectClaimAdapter) IsProjectClaimFake() (gcputil.OperationResult, error) {
```
NautiluX, line 240: The name suggests this was a function that returns some kind of a boolean, but it actually initiates all the fake processing, would probably be better to call it something like `EnsureProjectClaimFakeProcessed`?

pkg/controller/projectclaim/projectclaimadapter.go
```diff
@@ -169,7 +171,114 @@ func (c *ProjectClaimAdapter) FinalizeProjectClaim() (ObjectState, error) {
...
+			return gcputil.StopProcessing()
+		}
+		// Update fake project claim state
+		result, err = c.UpdateFakeProjectClaimState()
+		if err != nil {
+			return gcputil.RequeueWithError(operrors.Wrap(err, fmt.Sprintf("Could not update project claim specs for %s", c.projectClaim.Name)))
+		}
+		if !result {
+			return gcputil.StopProcessing()
+		}
```
NautiluX, line 274: While this is not necessary since we return anyways after the if block I guess you kept it for consistency with other function calls right?

pkg/controller/projectclaim/projectclaimadapter.go
```diff
@@ -169,7 +171,114 @@ func (c *ProjectClaimAdapter) FinalizeProjectClaim() (ObjectState, error) {
...
+		// Update fake project claim specs
+		result, err := c.UpdateFakeProjectClaimSpecs()
+		if err != nil {
+			return gcputil.RequeueWithError(operrors.Wrap(err, fmt.Sprintf("Could not update project claim specs for %s", c.projectClaim.Name)))
+		}
+		if !result {
+			return gcputil.StopProcessing()
+		}
+		// Update fake project claim state
```
NautiluX, line 267: This comment doesn't add much more context to the function call in the line below, so I think you could remove it.

PR #162, opened 2021-08-05
OSD-7839 - Remove wildcard from operator role

Tessg22:
### What type of PR is this? 
bug

GCP Project Operator uses a wildcard in the gcp.managed.openshift.io group but shouldn't. Removing wildcard from operator role.

Fixes https://issues.redhat.com/browse/OSD-7839

### Pre-checks:

### DIFF
--- a/deploy/cluster_role.yaml
+++ b/deploy/cluster_role.yaml
@@ -4,64 +4,66 @@ metadata:
   creationTimestamp: null
   name: gcp-project-operator
 rules:
-- apiGroups:
-  - ""
-  resources:
-  - pods
-  - services
-  - endpoints
-  - persistentvolumeclaims
-  - events
-  - configmaps
-  - secrets
-  verbs:
-  - create
-  - delete
-  - get
-  - list
-  - patch
-  - update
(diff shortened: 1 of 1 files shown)


deploy/cluster_role.yaml
```diff
@@ -55,7 +55,7 @@ rules:
 - apiGroups:
   - gcp.managed.openshift.io
   resources:
-  - '*'
+  - projectclaims
```
ramonbutter, line 58: do you think that is sufficient? Were you able to test this?
The status of a CR is listed under a different resource than the CR spec.
See f.e.:   https://github.com/openshift/route-monitor-operator/blob/20cc01df4b3024039c3c90748b263d27155ceab7/config/rbac/role.yaml#L83

Tessg22: $ oc auth can-i get projectclaims/status --as system:serviceaccount:gcp-project-operator:gcp-project-operator --as-group gcp.managed.openshift.io
yes

NautiluX (commented):
/lgtm

PR #169, opened 2022-02-20
Add API submodule

jhernand:
### What type of PR is this?

Refactor.

This is a refactor of the APIs package to make it a Go sub-module, so that other projects that use these APIs don't need to import all the transitive dependencies, specially the `controller-runtime` package and all its transitive dependencies.

This is related to [SDE-1586](https://issues.redhat.com/browse/SDE-1586) (OCM support for HyperShift) but it doesn't directly fixes it. It is a necessary to be able to import other API packages (MCE/ACM and HyperShift) that are otherwise incompatible with the versions required by this project.

No special notes.

It doesn't change any behavior or the versions of the dependencies, just the Go module structure.


pkg/gcpclient/client.go
```diff
@@ -19,7 +19,7 @@ import (
 	"google.golang.org/api/googleapi"
 	iam "google.golang.org/api/iam/v1"
 	"google.golang.org/api/option"
-	serviceManagment "google.golang.org/api/servicemanagement/v1"
+	"google.golang.org/api/serviceusage/v1"
```
georgettica, line 22: good to know the api has changed

georgettica:
Letting you know that on mobile or desktop the review process is really tough (doesn't handle it well)

georgettica:
@jhernand sad but ok :(

PR #170, opened 2022-02-21
Update dependencies

jhernand:
### What type of PR is this?

Refactor.

This patch updates the dependencies of the project. This is intended to
make it easier to import the API package from other projects. Currently
that is difficult because the code generated by the old version of the
`openapi-gen` tool isn't compatible with recent versions of the Kubernetes
API and the `controller-runtime` projects.

There are many changes in the `go.mod` and `go.sum` files, and in the
`vendor` directory. But the more relevant changes in the actual code of
the project are the following:

- Avoid importing the `operator-sdk` project. That isn't necessary with
  the new version of the Operator SDK. Applications that need components
  from the SDK can import the `operator-lib` project, but this application
  doesn't even need that.

- Add the `+listType=atomic` annotation to the API types. In the new
  version of the `openapi-gen` tool this is mandatory for fields that
  are slices.

- Add the context parameter to the `Reconcile` method of the
  reconcilers. This is required in the new version of the
  `controller-runtime` project.

- Use `client.Object` instead of `runtime.Object`. This is required in
  the new version of the `controller-runtime` project.

- Use the `serviceusage` package instead of `servicemanagement` to get
  and enable Google API services. This is required in the new version of
  the Google API project.

- Adapt log, leadership and metrics configuration to the new version of
  the `controller-runtime` package.

Unfortunately these changes can't be done in separate patches because
updating a single dependency pulls many other dependencies, and
introduces conflicts.

https://issues.redhat.com/browse/SDE-1586

It is backward compatible.


georgettica:
Do you need eyes on this? Seems like a massive PR.

Additionally, if you need help changing the CI in some where LMK and I'll explain it

jhernand:
Thaks @georgettica. Yes, I need eyes on this (and on #169, it is included here for convinience). Both are very convenient (or necessary) to make progress with HyperShift support in OCM.

This is massive because it includes all the changes to the vendor directory. I'll try to separate that to different commits to simplify your review. But all needs to go in the same pull request, otherwise build will just fail.

georgettica:
also, as I tried to point out, the CI (openshift/release) is broken.. did you find the culprit?

so @jhernand  assign me `/assign rogreen` when you want me to TAL

jhernand:
Yes, I noticed the CI failures, thanks @georgettica. I am trying to understand why it fails. I am looking at the logs of the jobs. I will assign to you (or ask for help) when I find the issue. In my local environment `make` and `make test` work fine. I guess that the CI environment is using different versions of the tools, but not sure yet.

Meanwhile I am assigning #169 to you.

jhernand:
Updated to include the installation of Go 1.17 as in #171.

PR #171, opened 2022-02-22
Replace Go 1.16 with 1.17 in builder image

jhernand:
### What type of PR is this? 

Refactor.

This patch changes the builder image so that it uses Go 1.17. Note that
there are currently no version of the _boilerplate_ base image that
contains Go 1.17, so the approach here is to remove Go 1.16 and install
1.17 explicitly.

Updating to Go 1.17 is necessary because without it it is not possible
to update to the latest versions of the Kubernetes API packages. In
particular the [sigs.k8s.io/json](https://github.com/kubernetes-sigs/json)
package doesn't work with Go 1.16 because it uses the `IsExported`
method that was added in Go 1.17.

This is a prerequisite for #170, which is a prerequisite for [SDA-5500](https://issues.redhat.com/browse/SDA-5500).

No special notes.

Backward compatible.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,4 +1,13 @@
 FROM quay.io/app-sre/boilerplate:image-v2.1.0 AS builder
+
+# Replace Go 1.16 with 1.17:
+RUN \
+    rm -rf /usr/local/go && \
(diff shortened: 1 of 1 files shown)


jhernand:
This doesn't help, see #170. I am closing.

PR #176, opened 2022-03-30
Remove vendor dir

cblecker:
### What type of PR is this? 
cleanup

Removes the unused vendor dir

_Fixes #_

### Pre-checks:


georgettica:
it is used I think.
are we moving away from the vendoring?

fahlmant:
Vendoring is no longer needed with go modules. We've moved away from it in all other repos we manage

georgettica:
When was this announced?
/lgtm
/approve
Then

fahlmant:
@georgettica This happened when go 1.16 was released in Feb 2021

PR #186, opened 2023-01-11
Pre-validate that service account credentials are usable before creating secret

mjlshen:
The main point of this change is this new function, which is used to validate that service account keys are valid before creating the associated Kubernetes secret. This is important because GCP's IAM is eventually consistent, worst case up to a few hours, so by validating this we can ensure the credentials are valid before kicking off an install, which would otherwise fail.

```go
// ValidateServiceAccountKey performs validation that a service account's keys can list at least 1 project to mirror the
// check that the installer does:
// https://github.com/openshift/installer/blob/ce13271664f492088048a2e90f208b8b51ea88ca/pkg/asset/installconfig/gcp/validation.go#L234-L248
//
// This check is necessary because GCP IAM is eventually consistent and in unfortunate times, eventually consistent
// after a few hours.
func ValidateServiceAccountKey(ctx context.Context, key *iam.ServiceAccountKey) error {
	creds, err := google.CredentialsFromJSON(ctx, []byte(key.PrivateKeyData), "https://www.googleapis.com/auth/cloud-platform")
	if err != nil {
		return fmt.Errorf("could not parse service account credentials: %w", err)
	}

	client, err := cloudresourcemanager.NewService(ctx, option.WithCredentials(creds))
	if err != nil {
		return fmt.Errorf("could not create cloudresourcemanager client with service account credentials: %w", err)
	}

	resp, err := client.Projects.List().Do()
	if err != nil {
		return fmt.Errorf("failed to list projects with service account credentials: %w", err)
	}

	if len(resp.Projects) == 0 {
		return errors.New("can view 0 projects with these service account credentials")
	}

	return nil
}
```

Fixes OSD-14445


rafael-azevedo (approved):
/lgtm

PR #194, opened 2023-06-06
[OSD-15800] Delete IAM policy bindings before deleting Service Account

zmird-r:
### What type of PR is this? 
Bug fix

To prevent policy bindings being left behind

Fixes [OSD-15800](https://issues.redhat.com/browse/OSD-15800)

### Pre-checks:

### DIFF
--- a/controllers/projectreference/projectreference_adapter.go
+++ b/controllers/projectreference/projectreference_adapter.go
@@ -468,9 +468,9 @@ func (r *ReferenceAdapter) deleteServiceAccount() error {
 
 	r.logger.V(1).Info("after get")
 
-    if err := r.DeleteIAMPolicy(sa.Email, 0); err != nil {
+	if err := r.DeleteIAMPolicy(sa.Email, 0); err != nil {
 		return operrors.Wrap(err, "could not delete the IAM policy, something happened")
-    }
+	}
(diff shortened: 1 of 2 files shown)


controllers/projectreference/projectreference_adapter.go
```diff
@@ -465,8 +465,13 @@ func (r *ReferenceAdapter) deleteServiceAccount() error {
 
 		return operrors.Wrap(err, "could not get the SA, something happened")
 	}
+
 	r.logger.V(1).Info("after get")
 
+	if err := r.DeleteIAMPolicy(sa.Email, 0); err != nil {
```
typeid, line 471: **nitpick:** use the text format instead of the numerical struct value.  
```suggestion
	if err := r.DeleteIAMPolicy(sa.Email, util.ServiceAccount); err != nil {
```

typeid (changes requested):
Small nitpick, rest looks good to me :)

controllers/projectreference/projectreference_adapter.go
zmird-r: Thank you!

NautiluX (commented):
/lgtm

PR #206, opened 2023-09-29
Add compute.networkAdmin to the osd-managed-admin SA when working wit…

jnpacker:
…h OSD GCP Shared VPC's

### What type of PR is this? 
Shared VPC with passThrough (required) needs the osd-managed-admin to have compute.networkAdmin role in the Service Acount.

Includes compute.networkAdmin on the osd-managed-admin in the Service Project.

_Fixes #_

### Pre-checks:

### DIFF
--- a/controllers/projectreference/projectreference_adapter.go
+++ b/controllers/projectreference/projectreference_adapter.go
@@ -86,6 +86,7 @@ var OSDSharedVPCRoles = []string{
 	"roles/iam.securityReviewer",
 	"roles/compute.loadBalancerAdmin",
 	"roles/resourcemanager.tagUser",
+	"roles/compute.networkAdmin",
 }
 
 // ReferenceAdapter is used to do all the processing of the ProjectReference type inside the reconcile loop


fahlmant:
@jnpacker Is there a Jira card specifically explaining why this permission is needed here? We need that for posterity and compliance tracking.

jnpacker:
Yes. https://issues.redhat.com/browse/SDE-3496

fahlmant:
Thanks! I moved it to the OSD project for tracking https://issues.redhat.com/browse/OSD-18803

I see the error specifically mentions `networksecurity.firewallEndpoints.use`. Is the full role `roles/compute.networkAdmin` required for that?

fahlmant:
Does networkAdmin contain permissions for other networksecurity resources, other than firewall?

jnpacker:
Yes. But it affects the user experience to ask them to build a custom Role. A bunch of the 

This is an OCP requirement that changed in 4.14, so I suspect they will be updating there documentation to support PassThrough mode on GCP.  Also the role only comes into play and is applied when the sharedVPC flag is used, further limiting.

The osd-managed-admin already has compute admin and a few other high level roles.
The user under normal circumstances in the Service Project is being granted:
Compute Admin
IAM Security Admin
Service ACcount Admin
Service Account Key Admin
Service ACcount User
Storage Admin
DNS Administrator
COmpute Load Balancer Admin
IAM ROle VIewer

Just to clarify, a Service Account is created with this role when PassThrough is NOT used (the default).  The difference is that because we're using passthrough, all permissions need to be on the top level SA (for the cluster in question) that is then used, where under normal circumstances, CCO creates an SA with this permission.

jnpacker:
Just as a double check, under the normal flow in OSD GCP for OCP 4.14, CCO will create an SA name: CLUSTER_NAME-master-node, this will have the Compute Network Admin permission, so that Ingress configuration will function correctly.  So we are NOT granting a role that isn't used as part of the normal GCP deployment, but because VPC Share requires PassThrough credentials, the primary SA needs to have this permission.  As mentioned above, this permission is only added when an SA is requested for SharedVPC as well, limiting the scope of the change.

PR #215, opened 2024-01-25
Fix log statement to print function name

bergmannf:
If the log level gets set to > 2 it will crash, because the logging function requires either 1 or an uneven number of arguments.

### What type of PR is this? 

_Fixes #_

### Pre-checks:


mjlshen (approved):
/lgtm

typeid (approved):
Thanks for the fix @bergmannf 
/approve

PR #216, opened 2024-01-31
Temporarily modify reconcile loop so that EnsureProjectConfigured will run

mjlshen:
### What type of PR is this? 
feature

We need to add an IAM role/rolebinding for existing osd-managed-admin serviceaccount that this operator creates. In the future, this serviceaccount will be managed by customers so that future permission changes in GCP can be actioned by the customer and not by us.

Also `go fmt`'d the codebase

[OSD-20579](https://issues.redhat.com//browse/OSD-20579)


rafael-azevedo (approved):
/lgtm

PR #230, opened 2024-09-11
feat(OSD-25439): Set up IAP permissions and api

typeid:
### What type of PR is this? 
feature

This PR adds the api and permissions needed for IAP (identity aware proxy) SRE access, which will be used to ssh to nodes for GCP PSC clusters. 

Fixes #https://issues.redhat.com/browse/OSD-25439

We don't need to reconcile all projects as PSC will only be for new clusters created after its release. 

For WIF, the enablement is done in the MCC wif template. 

### Pre-checks:


NautiluX (commented):
/lgtm

PR #245, opened 2025-07-14
SREP-102 Updates to prevent stuck subscriptions

dakotalongRH:
Update specs to prevent stuck subscriptions

https://issues.redhat.com/browse/SREP-102

### Pre-checks:

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -15,7 +15,7 @@
 package config
 
 const (
-	OperatorName      string = "gcp-project-operator"
-	OperatorNamespace string = "gcp-project-operator"
-	EnableOLMSkipRange string = "true"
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -17,4 +17,5 @@ package config
 const (
 	OperatorName      string = "gcp-project-operator"
 	OperatorNamespace string = "gcp-project-operator"
+	EnableOLMSkipRange string = "true"
```
joshbranham, line 20: This should fail linting, but either way can you align these? You can also just omit `string` and let the type system infer that. 

```golang
package config

const (
	OperatorName       = "gcp-project-operator"
	OperatorNamespace  = "gcp-project-operator"
	EnableOLMSkipRange = "true"
)
```

PR #291, opened 2026-03-20
Fix service account namespace in ClusterRoleBindings

eth1030:
Complete the namespace fix that was started in commit 9c79f64. That commit fixed the ServiceAccount namespace from gcp-project-operator-client to gcp-project-operator, but the ClusterRoleBindings were never updated to match.

This latent bug went unnoticed because OLM deployments auto-generate RBAC from the CSV and don't use these standalone ClusterRoleBinding files. However, when migrating to PKO (Package Operator), these manifests are used directly, exposing the inconsistency.

This caused RBAC permission errors in staging after PKO migration: "User 'system:serviceaccount:gcp-project-operator:gcp-project-operator-client' cannot get resource 'projectclaims'"

Changes:
- Fix deploy/uhc_cluster_role_binding.yaml namespace reference
- Fix deploy_pko/ClusterRoleBinding-gcp-project-operator-client.yaml namespace reference

Both now correctly reference the gcp-project-operator namespace where the ServiceAccount actually exists.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### What type of PR is this? 
bug

pko migration surfaced a bug 

part of olm -> pko migration: https://redhat.atlassian.net/browse/SREP-3558

### Pre-checks:

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -6,7 +6,7 @@ WORKDIR ${OPERATOR_PATH}
 # Build
 RUN make go-build
 
-FROM registry.access.redhat.com/ubi9/ubi-minimal:9.7-1773204619
+FROM registry.access.redhat.com/ubi9/ubi-minimal:9.7-1773939694
 ENV OPERATOR_PATH=/gcp-project-operator \
     OPERATOR_BIN=gcp-project-operator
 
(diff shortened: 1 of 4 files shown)


typeid (approved):
/lgtm
/approve

